PR #39, opened 2023-10-11
CV2-3435 add queue prefix

DGaffney:


DGaffney:
As per requirements on @sonoransun's end to have a way to differentiate queues between environments!

lib/helpers.py
```diff
@@ -5,7 +5,7 @@ def get_environment_setting(os_key: str) -> str:
     """
     Get environment variable helper. Could be augmented with credential store if/when necessary.
     """
-    return os.environ.get(os_key)
+    return os.environ.get(os_key, "")
```
skyemeedan, line 8: maybe add comment why it defaults to blank?

skyemeedan (approved):

PR #45, opened 2023-10-13
CV2-3852 fix queue prefix logic

DGaffney:
References 3852 in Jira - 

During the last deploy we had to rush in a fix to presto that would add a `QUEUE_PREFIX` environment variable so that we could have differently named SQS queues for the QA and Live environments. This came into conflict with the existing way we check for environment settings, where we opt to use a local value first by default. Our tests were broken as a result, since there's no local QUEUE_PREFIX supplied by default. I went ahead and added fixes to unify the calling for this `QUEUE_PREFIX` setting, and set it up to use the environment variable for it by default. This should fix the issues and clean up the incongruencies we had this week.

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -3,7 +3,10 @@ version: '3.9'
 services:
   app:
     platform: linux/amd64
-    build: .
+    build: 
+      context: .
+      args:
+        - PRESTO_PORT=${PRESTO_PORT}
     env_file:
       - ./.env_file
     depends_on:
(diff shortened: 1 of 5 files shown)


skyemeedan:
seems like this replacing hard-coded prefix substitution with the ability to grab from environment (can you add one line description to PR?)  you tested and it worked locally?

DGaffney:
> seems like this replacing hard-coded prefix substitution with the ability to grab from environment (can you add one line description to PR?) you tested and it worked locally?

writing the description currently, got pulled away from the computer after I pushed up

skyemeedan (approved):

PR #47, opened 2023-10-31
Cv2 3833 compute and save sscd embeddings for images

ahmednasserswe:


Dockerfile
```diff
@@ -8,6 +8,11 @@ EXPOSE ${PRESTO_PORT}
 WORKDIR /app
 ENV DEBIAN_FRONTEND=noninteractive
 
+RUN git clone https://github.com/facebookresearch/sscd-copy-detection.git
```
skyemeedan, line 11: maybe these should be installed after the system library stuff?

lib/model/image_sscd.py
```diff
@@ -0,0 +1,70 @@
...
+import requests
+import numpy as np
+
+class Model(Model):
+    def compute_sscd(self, image_url: str) -> str:
+        """Compute perceptual hash using ImageHash library
+        :param im: Numpy.ndarray
+        :returns: Imagehash.ImageHash
+        """
+        # pdq_hasher = PDQHasher()
```
skyemeedan, line 22: should these pdq_hash things still be here?

lib/model/image_sscd.py
```diff
@@ -0,0 +1,70 @@
...
+        # hash_and_qual = pdq_hasher.fromBufferedImage(iobytes)
+        # return hash_and_qual.getHash().dumpBitsFlat()
+        normalize = transforms.Normalize(
+            mean=[0.485, 0.456, 0.406], std=[0.229, 0.224, 0.225],
+        )
+        small_288 = transforms.Compose([
```
skyemeedan, line 28: can you add some comments here about what is going on? and/or link to paper or docs about the implementation?

lib/model/image_sscd.py
```diff
@@ -0,0 +1,70 @@
+from typing import Dict
+import io
+import urllib.request
+
+from lib.model.model import Model
+
+from pdqhashing.hasher.pdq_hasher import PDQHasher
```
skyemeedan, line 7: not using this import anymore?

Dockerfile
```diff
@@ -8,6 +8,11 @@ EXPOSE ${PRESTO_PORT}
...
+RUN cd sscd-copy-detection && python -m pip install -r ./requirements.txt --extra-index-url https://download.pytorch.org/whl/cu113
```
skyemeedan, line 12: Are we using the same docker file for all the models?  I think we need to have each one be able to have its own dependencies (so we don't install pytorch if not needed)

skyemeedan (commented):
Pretty cool to see a new model deployed! I think it would be really helpful to have a basic test for this?  Like 3 images, two are expected to have embeddings match and one that doesn't?

Dockerfile
computermacgyver: Agreed that we want to get to separate Dockerfiles for each model. We'll make that change next year.

computermacgyver (changes requested):

lib/model/image_sscd.py
computermacgyver: I've removed now.

lib/model/image_sscd.py
computermacgyver: @ahmednasserswe this is still to do

lib/model/image_sscd.py
computermacgyver: I've removed.

Dockerfile
computermacgyver: I've solved the build issues and moved the needed packages to `requirements.txt` where they belong @ahmednasserswe

lib/model/image_sscd.py
```diff
@@ -0,0 +1,70 @@
...
+            transforms.ToTensor(),
+            normalize,
+        ])
+        skew_320 = transforms.Compose([
+            transforms.Resize([320, 320]),
+            transforms.ToTensor(),
+            normalize,
+        ])
+
+        model = torch.jit.load("sscd_disc_mixup.torchscript.pt")
```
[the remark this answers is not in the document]
computermacgyver: I've done this now.

lib/model/image_sscd.py
```diff
@@ -0,0 +1,52 @@
...
+import torch
+from lib.logger import logger
+import numpy as np
+from PIL import Image
+import urllib.request
+
+class Model(GenericImageModel):
+    def __init__(self):
+        super().__init__()
+        #FIXME: Load from a Meedan S3 bucket
```
computermacgyver, line 16: This is still to do

computermacgyver:
Currently failing due to a Docker issue
```
#1 pulling image moby/buildkit:buildx-stable-1
#1 pulling image moby/buildkit:buildx-stable-1 5.7s done
#1 ERROR: Error response from daemon: Get "https://registry-1.docker.io/v2/": EOF
```

It's building and working fine locally because this image is cached. 

Before we merge this, we need to write tests for SSCD. Can you please look at `test/lib/model`test_image_pdq.py` and create similar tests for SSCD @ahmednasserswe ? :pray:

lib/model/image_sscd.py
computermacgyver: @ahmednasserswe Can you please add the comments requested here?

computermacgyver:
@DGaffney This was not merged before the refactor. Can you please review? We'd like to get this added to Presto after the next deploy.  The conflicts with image and test_image should be minor. Those files are just renamed to have `pdq` in them since `image` is too generic. We want the specific model names to be clear what models we're using as we change them.

DGaffney:
> @DGaffney This was not merged before the refactor. Can you please review? We'd like to get this added to Presto after the next deploy. The conflicts with image and test_image should be minor. Those files are just renamed to have `pdq` in them since `image` is too generic. We want the specific model names to be clear what models we're using as we change them.

I can take a look this week.

PR #48, opened 2023-11-03
CV2-2875 tweak response structure

DGaffney:
A ton of fixes on this ticket (which is just tied into the omnibus 2875 ticket) to change the response structure to be more consistent across modalities. This does some things that are *good*:

- It simplifies the schema into a singular schema, and actually *uses* that schema to reflect input and output messages,
- It makes the response packets from all modalities look basically the same in terms of their response structure, and stops embedding the response needlessly,

but it also does something I *don't* like and don't necessarily have a good answer for: 

The `GenericItem` is overly generic, and is not as nice as what I would prefer (which would be to have strict `TextItem` or `VideoItem` etc, modality-specific, schemas). The reason you *can't* do that is because, when you are loading the response data from SQS, at that point, you've sort of lost the thread in terms of what the type of object is. We'd need to build some sort of mapping of "this type of model uses this schema", and do that in a way that wasn't terrible and too loose, in order to pull it off (The crux of the issue is at `bodies = [schemas.Message(**json.loads(message.body)) for message, queue in messages_with_queues]` inside `lib/queue/processor.py` - if you look there, there's not really a great way to retrieve the model type that generated the object - we'd need to serialize that information alongside the object, and update all the code to reflect that. It's a bit of a meady thing). I'm not sure if I want to pull that into this ticket quite yet, but it would be required if we wanted to make the `GenericItem` thing go away. I welcome thoughts!


skyemeedan:
> we'd need to serialize that information alongside the object

DGaffney:
> > we'd need to serialize that information alongside the object

Sounds like you'd like to serialize that information alongside the object? Should we do that *now* or in a separate thing?

DGaffney:
Lets just do this fixing now - will come back later today with some updates...

skyemeedan (approved):

computermacgyver (approved):

PR #50, opened 2023-12-17
First attempt at tag based deploy workflow.

sonoransun:
This is a draft attempt at a tag based Live deployment workflow for Presto.

One remaining question mark is how to tie the image to the tag, as we currently have `master` and specific revision `sha`s as identifier options; no tag names.

We could just deploy master when tagged, as it should always be the latest that we deploy? But that's hacky and will have unintended side effects at some point in the future. I'll look at this again once I've got a bit more brain to apply...

### DIFF
--- a/.github/workflows/ci-tag-deploy.yml
+++ b/.github/workflows/ci-tag-deploy.yml
@@ -1,0 +1,87 @@
+name: Live Deploy
+
+on:
+  push:
+    # Pattern matched against refs/tags
+    tags:        
+      - '*'
+
+permissions:
+  id-token: write
+  contents: read
+
+jobs:
+  live-deploy:
+    runs-on: ubuntu-latest
+
+    steps:
+    - name: Kick off Terraform deploy in sysops/
+      id: sysops-deploy
+      run: |
+        curl \
+          -X POST \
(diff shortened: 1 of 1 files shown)


.github/workflows/ci-tag-deploy.yml
```diff
@@ -0,0 +1,87 @@
...
+    steps:
+    - name: Kick off Terraform deploy in sysops/
+      id: sysops-deploy
+      run: |
+        curl \
+          -X POST \
+          -H "Accept: application/vnd.github+json" \
+          -H "Authorization: Bearer ${{ secrets.SYSOPS_RW_GITHUB_TOKEN }}" \
+          -H "X-GitHub-Api-Version: 2022-11-28" \
+          https://api.github.com/repos/meedan/sysops/actions/workflows/deploy_${{ github.event.repository.name }}.yml/dispatches \
```
skyemeedan, line 26: Does this trigger script to look at the tagged repository and redeploy all of the associated services?   Are we assuming that the workflow is "merge to main, run in QA for a while, then tag to trigger live release?" i.e. the images to be deployed are already build and pushed to ECR?

PR #53, opened 2024-01-10
CV2-3700-sscd-image-similarity-algorithm

ahmednasserswe:
Implement support for SSCD-copy-detection algorithm


computermacgyver (changes requested):
Please go back through https://github.com/meedan/presto/pull/47/files @ahmednasserswe and pull in the associated changes that are needed including changes to `.gitignore`  and `requirements.txt` and the test for sscd. Have you tested this locally?

lib/model/image_sscd.py
```diff
@@ -0,0 +1,58 @@
...
+
+class Model(Model):
+    def __init__(self):
+        super().__init__()
+        #FIXME: Load from a Meedan S3 bucket
+        try:
+            self.model = torch.jit.load("sscd_disc_mixup.torchscript.pt")
+        except:
+            logger.info("Downloading SSCD model...")
+            m=urllib.request.urlopen("https://dl.fbaipublicfiles.com/sscd-copy-detection/sscd_disc_mixup.torchscript.pt").read()
```
DGaffney, line 22: Please make all these strings constants and/or environment variables. Also - @computermacgyver do we want to host this in our own infrastrcture?

lib/model/image_sscd.py
```diff
@@ -0,0 +1,58 @@
...
+
+    def compute_sscd(self, iobytes: io.BytesIO) -> str:
+        """Compute perceptual hash using ImageHash library
+        :param im: Numpy.ndarray #FIXME
+        :returns: Imagehash.ImageHash #FIXME
+        """
+        # from SSCD-copy-detection readme https://github.com/facebookresearch/sscd-copy-detection/tree/main#preprocessing
+        # Normalization using the mean and std of Imagenet
+        normalize = transforms.Normalize(
+            mean=[0.485, 0.456, 0.406], std=[0.229, 0.224, 0.225],
```
DGaffney, line 36: Please move these to constants.

DGaffney (changes requested):
Can we also get an end-to-end test that confirms this interfaces well with queues?

.idea/.DS_Store
computermacgyver, line 1: Please remove this file @ahmednasserswe

lib/model/image_sscd.py
computermacgyver: I'm OK with keeping the URL pointing to fbaipublicfiles.com for now. I have a backup in our own S3 bucket in case we need it at some point.

computermacgyver (changes requested):

lib/model/image_sscd.py
DGaffney: No need to create a separate constants file, @ahmednasserswe - I think if you just put these up top with a descriptive name that'd be sufficient!

lib/model/image_sscd.py
```diff
@@ -0,0 +1,60 @@
...
+from lib import schemas
+from torchvision import transforms
+import torch
+from lib.logger import logger
+import numpy as np
+from PIL import Image
+import urllib.request
+
+imagenet_normalization_mean = [0.485, 0.456, 0.406]
+imagenet_normalization_std = [0.229, 0.224, 0.225]
```
DGaffney, line 15: Upcase these please

lib/model/image_sscd.py
DGaffney: Where did we land with running this within the existing image model? Do we still need to refactor for that?

lib/model/image_sscd.py
```diff
@@ -0,0 +1,60 @@
...
+class Model(Model):
+    def __init__(self):
+        super().__init__()
+        #FIXME: Load from a Meedan S3 bucket
```
DGaffney, line 19: Is this comment valid? Do we need to do something else here or delete this line?

lib/model/image_sscd.py
```diff
@@ -0,0 +1,60 @@
...
+        try:
+            self.model = torch.jit.load("sscd_disc_mixup.torchscript.pt")
+        except:
```
DGaffney, line 22: Can we get a more explicit/focused exception here? What are we rescuing against in this situation?

lib/model/image_sscd.py
```diff
@@ -0,0 +1,60 @@
...
+
+    def compute_sscd(self, iobytes: io.BytesIO) -> str:
+        """Compute perceptual hash using ImageHash library
+        :param im: Numpy.ndarray #FIXME
+        :returns: Imagehash.ImageHash #FIXME
+        """
+        # from SSCD-copy-detection readme https://github.com/facebookresearch/sscd-copy-detection/tree/main#preprocessing
+        # Normalization using the mean and std of Imagenet
+        normalize = transforms.Normalize(
+            mean = imagenet_normalization_mean , std = imagenet_normalization_std,
```
DGaffney, line 38: Upcase these here when you upcase above

DGaffney (changes requested):

lib/model/image_sscd.py
ahmednasserswe: @computermacgyver what do you think? there has been an unresolved conversation about putting the file public in the s3 bucket and the its rules. Should we reopen this conversation and see how to handle hosting the file on the s3 bucket and evantually refer to the file location here? or should we just keep it in the s3 bucket as a backup and we use it in case it was removed from its current url?

DGaffney: Waiting for @computermacgyver to weigh in here!

computermacgyver:
Closing this in favor of #59

PR #54, opened 2024-01-12
Cv2 4126 video

DGaffney:


lib/s3.py
```diff
@@ -1,12 +1,37 @@
...
 import boto3
+from botocore.client import Config
+from botocore.exceptions import ClientError
 from lib.logger import logger
 def upload_file_to_s3(bucket: str, filename: str):
     """
     Generic upload helper for s3. Could be moved over to helpers folder...
     """
+    s3_url = os.getenv('S3_ENDPOINT')
+    access_key = os.getenv('AWS_ACCESS_KEY_ID')
```
skyemeedan, line 11: In timpani, we did access using the roles, vs access key, so maybe this is not needed?

lib/s3.py
```diff
@@ -1,12 +1,37 @@
...
+        s3_client.head_bucket(Bucket=bucket)
+    except ClientError as e:
+        if e.response['Error']['Code'] == 'NoSuchBucket' or int(e.response['Error']['Code']) == 404:
+            # Create the bucket since it does not exist
+            s3_client.create_bucket(Bucket=bucket)
+            logger.info(f'Created bucket {bucket} in MinIO.')
+        else:
+            # Other errors like permissions issues
+            logger.error(f'Error in accessing bucket {bucket}: {e}, {bucket} {filename}')
+            return
```
skyemeedan, line 34: shouldn't this still raise the error?

skyemeedan (approved):

lib/s3.py
DGaffney: I'll let @sonoransun weigh in on this - this is how I've _usually_ done this sort of thing, but maybe its not needed?

lib/s3.py
DGaffney: Fair enough - I'll raise

lib/s3.py
DGaffney: FWIW I would _prefer_ leaving it this way as currently written as it interfaces nicely with minio locally - no biggie if that's a problem though!

PR #55, opened 2024-01-31
DEVOPS-528 Master workflow trigger for Live deployments

sonoransun:
This change adds a workflow triggered by tag events on the presto repository. This is used to invoke Live deployments via the sysops Terraform workflow for presto.

Note that this will also need a corresponding change in sysops to use the qa or live workspace as determined by a new workflow event data element `workspace`.


.github/workflows/ci-live-tag-deploy.yml
dmou, line 1: do we really need an entire separate workflow? this exposes our environments to anyone looking at the repository

dmou (changes requested):

.github/workflows/ci-deploy.yml
```diff
@@ -77,19 +77,16 @@ jobs:
...
-          curl \
+        curl \
           -X POST \
           -H "Accept: application/vnd.github+json" \
           -H "Authorization: Bearer ${{ secrets.SYSOPS_RW_GITHUB_TOKEN }}" \
           -H "X-GitHub-Api-Version: 2022-11-28" \
           https://api.github.com/repos/meedan/sysops/actions/workflows/deploy_${{ github.event.repository.name }}.yml/dispatches \
-          -d '{"ref": "master", "inputs": {"git_sha": "${{ github.sha }}", "branch": "'$BRANCH'"}}'
-        fi
+          -d '{"ref": "master", "inputs": {"git_sha": "${{ github.sha }}", "workspace": "qa"}}'
```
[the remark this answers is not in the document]
sonoransun: How would we keep the mapping in sysops itself? (When sysops is fired I only get a commit SHA otherwise, and I can't tell from that which environment is intended.)

I might be missing a trick to knowing context for a workflow dispatch...

.github/workflows/ci-live-tag-deploy.yml
```diff
@@ -0,0 +1,155 @@
...
+      run: |
+        curl \
+          -X POST \
+          -H "Accept: application/vnd.github+json" \
+          -H "Authorization: Bearer ${{ secrets.SYSOPS_RW_GITHUB_TOKEN }}" \
+          -H "X-GitHub-Api-Version: 2022-11-28" \
+          https://api.github.com/repos/meedan/sysops/actions/workflows/deploy_${{ github.event.repository.name }}.yml/dispatches \
+          -d '{"ref": "master", "inputs": {"git_sha": "${{ github.sha }}", "workspace": "live"}}'
+
+    - name: Send GitHub Action trigger data to Slack workflow on success
```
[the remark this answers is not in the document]
sonoransun: Great; I had wondered about these! I remove the notifications from both deployment workflows.

.github/workflows/ci-live-tag-deploy.yml
sonoransun: I'm not sure how to get a commit ID from a tag, so we build the tag and get an image ID to use in deployments.

If there is a way to know the image ID associated with a QA deployment from a merge, that is equivalent to the applied tag, that would work! But I can't find a way to do this in a workflow fired by either event. Hence the different workflows for the different trigger events...

I'll keep looking for a better option.

dmou: https://docs.github.com/en/webhooks/webhook-events-and-payloads#push

image ID as in ecr image tag? it's whatever you tell it it is, which is what's defined in the deploy workflow. the ecr image you deploy to any environment should be specified by the git sha, which is always included as part of github events

.github/workflows/ci-deploy.yml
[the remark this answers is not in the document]
dmou: you can pass any variables you want to a dispatch, this is in the gha workflow docs

.github/workflows/ci-live-tag-deploy.yml
sonoransun: Yeah the commit_id / SHA - and our `container_image_tag`.

If I understand you correctly you're saying I can simplify things by doing the following:

1.  I can re-use the same deploy workflow but add a trigger for the appropriate `v*` syntax.
2. When a tag is applied it will reference a git_commit_sha in the workflow that is the same as when `master` was merged. In other words, I don't need to map from a tag to a git_commit_sha, because the tag event will reference the right merge git_commit_sha. (I didn't know if this was the case, and I actually assumed it was not due to some prior issues I've had with Travis triggers)
3. To avoid putting workspace details in the presto workflow, have it call the sysops deployment with either a git_commit_sha or a tag? And if a tag we deploy to Live instead? Or did you have a different idiom in mind for the presto vs. sysops sides of this?

dmou: 1. yep!
2. a tag is a git ref. if you take a look at the link, it provides attributes, several of which reference the sha of the commit a tag is pointing to. tags have nothing to do with branches in git, it won't track when master was merged. tagging off of only master is a practice we can choose to have, though. pushing a tag will change the git ref to update to whatever commit you're pointing it to next.
3. if it was a push to master branch event, it's in qa. if it's a tag push event, it goes to live. are there situations where we'd want other cases for those environments?

sonoransun: That makes sense. I think I've got the changes as mentioned working in a way that uses the same workflow for either merge or tag, invokes sysops deploy with an agnostic reference (master or tag for existing ref) so it knows QA or Live, and cleans out the Slack cruft.

I also updated the sysops pull request (https://github.com/meedan/sysops/pull/268) side of this so that it has a distinct workspace run rule for either merge or tag as mentioned.

I think this is getting close!

.github/workflows/ci-deploy.yml
```diff
@@ -77,78 +79,29 @@ jobs:
           --tag $ECR_REGISTRY/$ECR_REPOSITORY:$BRANCH \
           --file ./Dockerfile ./
 
-    - name: Kick off Terraform deploy in sysops/ if not Live environment
+    - name: Kick off Terraform deploy in sysops/
       id: sysops-deploy
+      if: ${{ github.event.pull_request.merged }}
```
dmou, line 84: there will never be a pull_request event, this deploy is triggered on push events

.github/workflows/ci-deploy.yml
```diff
@@ -77,78 +79,29 @@ jobs:
...
-                ]
-              }
-            ]
-          }
-      env:
-        SLACK_WEBHOOK_URL: ${{ secrets.ITS_BOTS_SLACK_WEBHOOK_URL }}
-        SLACK_WEBHOOK_TYPE: INCOMING_WEBHOOK
+    - name: Kick off Terraform deploy in sysops/
+      id: sysops-deploy
+      if: github.event_name == 'create' && startsWith(github.ref, 'refs/tags/v')
```
dmou, line 96: you might want to double check the difference between create events and push events, this could work with specific process controls but they are separate events

.github/workflows/ci-deploy.yml
```diff
@@ -77,78 +79,29 @@ jobs:
...
+      run: |
+        curl \
+          -X POST \
+          -H "Accept: application/vnd.github+json" \
+          -H "Authorization: Bearer ${{ secrets.SYSOPS_RW_GITHUB_TOKEN }}" \
+          -H "X-GitHub-Api-Version: 2022-11-28" \
+          https://api.github.com/repos/meedan/sysops/actions/workflows/deploy_${{ github.event.repository.name }}.yml/dispatches \
+          -d '{"ref": "tag", "inputs": {"git_sha": "${{ github.sha }}"}}'
```
dmou, line 104: this will run a workflow off of the `tag` branch

dmou (dismissed):

.github/workflows/ci-deploy.yml
```diff
@@ -77,78 +79,29 @@ jobs:
...
-                    }
-                  }
-                ]
-              }
-            ]
-          }
-      env:
-        SLACK_WEBHOOK_URL: ${{ secrets.CHECK_DEV_BOTS_SLACK_WEBHOOK_URL }}
-        SLACK_WEBHOOK_TYPE: INCOMING_WEBHOOK
+          -d '{"ref": "master", "inputs": {"git_sha": "${{ github.sha }}", "type": "push"}}'
```
skyemeedan, line 92: not sure if this is refering to the github side, or the source repository side.  But if it is on the source repo side some will be "main" instead of "master"

dmou: thanks skye! this is on the terraform deployment workflow repo side, so we are going off of master this time

.github/workflows/ci-deploy.yml
sonoransun: Ok, I think `if: github.event_name == 'create' && startsWith(github.ref, 'refs/heads')` will work, but it may actually fire too often? (No harm, just an extra deploy of master which is already deployed?)

I wonder if this should be  `...startsWith(github.ref, 'refs/heads/master')`

dmou: couple things: one, you haven't added a trigger for `create` events to this workflow either so this will never be a true statement. two, if this is for deployments off the main branch then you already have a push event. there's only two possible paths to trigger a deployment and they are distinct.

sonoransun: I do wish we had started with a `main` branch on presto...

PR #57, opened 2024-02-06
CV2-4247 Add timeout functionality

DGaffney:


lib/queue/worker.py
```diff
@@ -1,10 +1,13 @@
...
+from concurrent.futures import ThreadPoolExecutor, TimeoutError
 import json
-from typing import List
+from typing import List, Tuple
 from lib import schemas
 from lib.logger import logger
 from lib.queue.queue import Queue
 from lib.model.model import Model
 from lib.helpers import get_setting
+TIMEOUT = int(os.getenv("WORK_TIMEOUT", "60"))
```
skyemeedan, line 10: add comment to indicate units?  (seconds?)

lib/queue/worker.py
```diff
@@ -48,13 +51,68 @@ def safely_respond(self, model: Model) -> List[schemas.Message]:
...
+        """
+        return [schemas.Message(**{**json.loads(message.body), **{"model_name": model.model_name}})
+                for message, queue in messages_with_queues]
+
+    @staticmethod
+    def execute_with_timeout(func, args, timeout_seconds: int) -> List[schemas.Message]:
+        """
+        Executes a given hasher/fingerprinter with a specified timeout. If the hasher/fingerprinter execution time exceeds the timeout,
+        logs an error and returns an empty list.
+
```
skyemeedan, line 80: Will the job that timed out be killed, or does it keep running?  If it is still running, and a second attempt is made, does this mean the callback may be executed twice?

lib/queue/worker.py
```diff
@@ -48,13 +51,68 @@ def safely_respond(self, model: Model) -> List[schemas.Message]:
...
+        Parameters:
+        - func (callable): The function to execute.
+        - args (any): The arguments to pass to the function.
+        - timeout_seconds (int): The maximum number of seconds to wait for the function to complete.
+
+        Returns:
+        - List[schemas.Message]: The result of the function if it completes within the timeout, otherwise an empty list.
```
skyemeedan, line 87: How will the caller distinguish a job that timed out (probably requiring a retry) from a job that successfully returned no results?

lib/queue/worker.py
```diff
@@ -48,13 +51,68 @@ def safely_respond(self, model: Model) -> List[schemas.Message]:
...
-            try:
-                responses = model.respond([schemas.Message(**{**json.loads(message.body), **{"model_name": model.model_name}}) for message, queue in messages_with_queues])
-            except Exception as e:
-                logger.error(e)
-            self.delete_messages(messages_with_queues)
+        if not messages_with_queues:
+            return []
+        messages = self.extract_messages(messages_with_queues, model)
+        responses = self.execute_with_timeout(model.respond, messages, timeout_seconds=TIMEOUT)
+        self.delete_processed_messages(messages_with_queues)
```
skyemeedan, line 58: I think we should we check if the item was successfully processed and if not maybe retry, or send to 'dead letter' queue so that it isn't lost?

skyemeedan (commented):
How does timeout fits into the overall architecture (i.e. will work be lost? who should retry?  Do we already have corresponding timeouts on the blocking processes so they won't wait forever on a job that keeps timing out?)

I'm also generally worried about the amount of work it will take to iron out all of the edge cases of a work dispatch system.  Are there any pre-existing stable libraries or tools (Celery?) we could use to manage this?

lib/queue/worker.py
DGaffney: Ideally it is killed - at this point we have very little control - if it's hanging indefinitely, we have to try to kill it, but if it doesn't die, someone else may take the work up potentially depending on our design here - the good thing is I strain to think of a situation where running twice is dangerous?

lib/queue/worker.py
DGaffney: Fair enough!

lib/queue/worker.py
DGaffney: Good call - this is the point where we can figure out if we can safely delete the tasks

lib/queue/worker.py
DGaffney: Agreed!

lib/queue/worker.py
skyemeedan: maybe if it times out it can try to kill the process it started?  
> situation where running twice is dangerous
if it responds twice to an async request, couldn't that lead to multiple responses to user?

skyemeedan (approved):
I still think we need to design the retry strategy.  Like if something is taking more than a minute to process, we will give up and not mark as success.  So it it will be retried by another worker (yay!)  but if the reason it took more than a minute is something wrong with the content, eventually all the workers will by cycling retries of unfinishable tasks?  So need a retry counter?

DGaffney:
> I still think we need to design the retry strategy. Like if something is taking more than a minute to process, we will give up and not mark as success. So it it will be retried by another worker (yay!) but if the reason it took more than a minute is something wrong with the content, eventually all the workers will by cycling retries of unfinishable tasks? So need a retry counter?

I agree that we need a bit more color on the retry - I think lets ship this now to stem an even worse situation, but talk about this on the next ML call to decide on what the retry strategy looks like?

computermacgyver:
Looks like we currently simply log tasks that time out and do not retry? I'm fine with shipping that first and then following up with a retry strategy in another ticket. Particularly because I believe Check API already has a retry strategy. Ideally Alegre will also throw back an error to Check API when Presto errors so that Check API knows to retry (and it has a max number of times to retry already).

PR #59, opened 2024-02-08
Cv2 4242 implement sscd in image.py alongside pdq and create a new branch for it

ahmednasserswe:
implement both sscd and pdq together in image.py


lib/model/image.py
```diff
@@ -7,7 +7,27 @@
...
+from lib.logger import logger
+import numpy as np
+from PIL import Image
+
+IMAGENET_NORMALIZATION_MEAN = [0.485, 0.456, 0.406]
+IMAGENET_NORMALIZATION_STD = [0.229, 0.224, 0.225]
 class Model(Model):
+    def __init__(self):
+        super().__init__()
+        #FIXME: Load from a Meedan S3 bucket
```
DGaffney, line 21: Still needed?

lib/model/image.py
```diff
@@ -17,6 +37,35 @@ def compute_pdq(self, iobytes: io.BytesIO) -> str:
...
+            mean = IMAGENET_NORMALIZATION_MEAN , std = IMAGENET_NORMALIZATION_STD,
+        )
+        # It is recommended by publishers of SSCD-copy-detection to preprocess images for inference either resizing the small edge to 288 or resizing the image to a square tensor.
+        # resizing the image to a square tensor is more effecient on gpus but can lead to skewed images and so loss of information. So, we are resizing the small edge to 288
+        small_288 = transforms.Compose([
+            transforms.Resize(288),
+            transforms.ToTensor(),
+            normalize,
+        ])
+        # Keeping the code example of resizing the image to a square tensor
```
DGaffney, line 57: Drop these lines if we are never doing this?

lib/model/image_sscd.py
```diff
@@ -0,0 +1,60 @@
+from typing import Dict
```
DGaffney, line 1: Is there a compelling reason to persist a separate SSCD file @computermacgyver or @ahmednasserswe ?

lib/schemas.py
```diff
@@ -25,6 +28,8 @@ def set_body(cls, values):
         model_name = values.get("model_name")
         if model_name == "video__Model":
             values["body"] = VideoItem(**values["body"]).dict()
-        if model_name in ["audio__Model", "image__Model", "fptg__Model", "indian_sbert__Model", "mean_tokens__Model", "fasttext__Model"]:
+        if model_name == "image__Model":
+            values["body"] = ImageItem(**values["body"]).dict()
+        if model_name in ["audio__Model", "fptg__Model", "indian_sbert__Model", "mean_tokens__Model", "fasttext__Model"]:
```
DGaffney, line 33: I think this should be an elif?

DGaffney (changes requested):
Just a few points to check in on!

lib/model/image_sscd.py
computermacgyver: No I think this file can and should be removed

lib/model/image.py
```diff
@@ -7,7 +7,27 @@
...
+from PIL import Image
+
+IMAGENET_NORMALIZATION_MEAN = [0.485, 0.456, 0.406]
+IMAGENET_NORMALIZATION_STD = [0.229, 0.224, 0.225]
 class Model(Model):
+    def __init__(self):
+        super().__init__()
+        #FIXME: Load from a Meedan S3 bucket
+        try:
+            self.model = torch.jit.load("sscd_disc_mixup.torchscript.pt")
```
computermacgyver, line 23: Instead of the generic self.model, we should redact to something clearer like self.sscd_model

lib/model/image.py
```diff
@@ -30,8 +79,10 @@ def get_iobytes_for_image(self, image: schemas.Message) -> io.BytesIO:
...
 
-    def process(self, image: schemas.Message) -> schemas.GenericItem:
+
+    def process(self, image: schemas.Message) -> schemas.ImageItem:
         """
         Generic function for returning the actual response.
         """
-        return self.compute_pdq(self.get_iobytes_for_image(image))
+        img_iobytes = self.get_iobytes_for_image(image)
+        return {"hash_value": self.compute_pdq(img_iobytes), "sscd_value": self.compute_sscd(img_iobytes)}
```
computermacgyver, line 88: Let's put try-except error catching around compute_pdq and compute_sscd . If one of the two fails, I'd like to still be able to return the other

lib/model/image.py
computermacgyver: This should maybe be written as "#FUTURE: " . I think we don't need to change right now.

lib/model/image.py
computermacgyver: Ahmed and I discussed putting some logging into this function to track how much time is spent on SSCD and PDQ. We'll also pre-test on co-lab to have an approximate idea so we know how much additional work we're taking on by running SSCD for every image request

```

pdq=None
try:
	pdq_time=Time.now()
	pdq=self.compute_pdq(img_iobytes)
except:
	...
finally:
	pdq_time=Time.now()-pdq_time
	log....(f"PDQ took {pdq_time} seconds to process {filename}")
	

#Similar for sscd=self.compute_sscd(img_iobytes)

return {"hash_value": pdq, "sscd_value": sscd}
```

ahmednasserswe: I added a generic except as I can't tell what sort of errors might happen during pdq or sscd calculation. I also calculated the processing time in the `try` clause not `finally` as if the calculation failed, we don't need to log the processing time.

skyemeedan:
Closing this PR because when we move forward with this we will need to port changes to a new PR to align with refactoring

PR #60, opened 2024-02-21
CV2-4297 add suffix

DGaffney:


sonoransun (commented):
Perfect! This should do it 👍

PR #65, opened 2024-02-26
Ensure that fifo suffix comes at end of output queue names.

sonoransun:
This bugfix ensures that the `.fifo` suffix is appended to the end of output queues as well as input queue names.

There are no security concerns with this change that I can conceive of.

### DIFF
--- a/lib/queue/queue.py
+++ b/lib/queue/queue.py
@@ -109,7 +109,9 @@ class Queue:
         If output_queue_name was empty or None, set name for queue.
         """
         if not output_queue_name:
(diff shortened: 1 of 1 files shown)


sonoransun:
@computermacgyver , @skyemeedan : Devin is busy and I'm hoping one of you could take a look at this pull request. It's not elegant as fixes go. If we were on a newer Python version I could use the `removesuffix()` method, alas...

Let me know if you have any suggestions! Thanks :)

DGaffney (approved):

PR #68, opened 2024-02-28
Cv2 4136 timpani accessible yake keyword model

ahmednasserswe:
Creating model Yake keywords to be accessible by Timpani 
In reference to https://github.com/LIAAD/yake
References: [CV2-4136](https://meedan.atlassian.net/browse/CV2-4136)

[CV2-4136]: https://meedan.atlassian.net/browse/CV2-4136?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


lib/model/yake.py
```diff
@@ -0,0 +1,51 @@
...
+import io
+import urllib.request
+
+from lib.model.model import Model
+
+from lib import schemas
+
+from lib.logger import logger
+
+import yake
```
computermacgyver, line 11: This won't work. You cannot call the file `yake.py` and then `import yake` . The file must be called something different

lib/model/yake.py
```diff
@@ -0,0 +1,51 @@
...
+        :param windowSize: int
+        :param numOfKeywords: int
+        :returns: str
+        """
+        custom_kw_extractor = yake.KeywordExtractor(lan=language, n=max_ngram_size, dedupLim=deduplication_threshold,
+                                                    dedupFunc=deduplication_algo, windowsSize=windowSize,
+                                                    top=numOfKeywords, features=None)
+        keywords = custom_kw_extractor.extract_keywords(text)
+
+        return keywords
```
computermacgyver, line 36: Please refine this so that the return value is a dictionary. E.g.,
```
return { "keywords": keywords}
```

lib/schemas.py
```diff
@@ -27,4 +36,6 @@ def set_body(cls, values):
             values["body"] = VideoItem(**values["body"]).dict()
         if model_name in ["audio__Model", "image__Model", "fptg__Model", "indian_sbert__Model", "mean_tokens__Model", "fasttext__Model"]:
             values["body"] = MediaItem(**values["body"]).dict()
+        if model_name == "yake__Model":
```
computermacgyver, line 39: `elif`

computermacgyver (changes requested):
A few minor notes. Please add tests.

lib/schemas.py
```diff
@@ -16,6 +17,14 @@ class VideoItem(MediaItem):
     folder: Optional[str] = None
     filepath: Optional[str] = None
 
+class YakeItem(GenericItem):
```
computermacgyver, line 20: This appears to be wrong. This should be what is returned after running Yake, i.e., `{"keywords": [["a", 0.98], "b",0.8]]}`

lib/model/yake.py
```diff
@@ -0,0 +1,51 @@
...
+
+    def process(self, text: schemas.Message) -> schemas.GenericItem:
```
computermacgyver, line 38: Return type here should be `YakeItem` I believe

skyemeedan (changes requested):
* We need a test file that both spins up the model and sends and recieves request,  and also serves to pin down an example of input and output data structures
* Likely there is more supporting infrastructure needed for presto? (queues, etc?)

test/lib/model/test_yake_keywords.py
```diff
@@ -0,0 +1,68 @@
...
+            "body": {
+                "id": "1234",
+                "text": "Some Text",
+            },
+            "model_name": "yake_keywords__Model"
+        })
+        mock_yake_response.return_value = [["ball", 0.23]]
+        self.assertEqual(self.yake_model.process(message), {"keywords": [["ball", 0.23]]})
+
+    @patch('yake.KeywordExtractor.extract_keywords')
```
skyemeedan, line 23: This looks too me like we are patching the return value and not actually calling the model that we are trying to test?

test/lib/model/test_yake_keywords.py
```diff
@@ -32,6 +32,17 @@ def test_run_yake(self, mock_yake_response):
...
+    def test_run_yake_real(self):
+        message = schemas.Message(**{
+            "body": {
+                "id": "1234",
+                "text": "I love Meedan",
+            },
+            "model_name": "yake_keywords__Model"
+        })
+        results = self.yake_model.run_yake(**self.yake_model.get_params(message))
+        self.assertEqual(results, {"keywords": [('love Meedan', 0.0013670273525686505)]})
```
skyemeedan, line 44: Nice! looks like we submitted an id, but it didn't come back out of the model?  Or is presto going to handle the linking of request id payload to the keywords returned?
Also, can we use a more complex realistic example where we will get multiple keywords back?

ahmednasserswe: `process` function in `yake_keywords.py` returns GenericType which contains the id as well. This test case is for `run_yake`
I'll commit a longer test case

computermacgyver (approved):

DGaffney (approved):
Ship it!

caiosba (changes requested):
Please be sure the code here doesn't break CI: https://app.travis-ci.com/github/meedan/check-web/jobs/618966497

PR #78, opened 2024-03-07
Add a location constraint for bucket creation.

sonoransun:
The only region where a location constraint is not required is us-east-1.

### DIFF
--- a/lib/s3.py
+++ b/lib/s3.py
@@ -21,7 +21,7 @@ def upload_file_to_s3(bucket: str, filename: str):
     except ClientError as e:
         if e.response['Error']['Code'] == 'NoSuchBucket' or int(e.response['Error']['Code']) in [403, 404]:
             # Create the bucket since it does not exist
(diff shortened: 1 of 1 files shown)


lib/s3.py
```diff
@@ -21,7 +21,7 @@ def upload_file_to_s3(bucket: str, filename: str):
     except ClientError as e:
         if e.response['Error']['Code'] == 'NoSuchBucket' or int(e.response['Error']['Code']) in [403, 404]:
             # Create the bucket since it does not exist
-            s3_client.create_bucket(Bucket=bucket)
+            s3_client.create_bucket(Bucket=bucket, CreateBucketConfiguration={'LocationConstraint': region})
```
skyemeedan, line 24: maybe we should set this to default to `eu-west-1` if the paramter is missing in the env?  I think that would make line 11 `region = os.getenv('AWS_DEFAULT_REGION', 'eu-west-1')`

PR #85, opened 2024-05-21
Cv2 4606 delete message

DGaffney:
If we don't delete the existing message this code just has the net effect of *compounding the same task repeatedly* which is definitely not what we want! Increment the count, delete the old job, and add a new job.

### DIFF
--- a/lib/queue/worker.py
+++ b/lib/queue/worker.py
@@ -144,4 +144,5 @@ class QueueWorker(Queue):
             else:
                 updated_message = schemas.parse_message(message_body)
                 updated_message.retry_count = retry_count
+                queue.delete_messages(Entries=[self.delete_message_entry(message)])
                 self.push_message(self.input_queue_name, updated_message)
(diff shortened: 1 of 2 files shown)


lib/queue/worker.py
```diff
@@ -144,4 +144,5 @@ def increment_message_error_counts(self, messages_with_queues: List[Tuple]):
             else:
                 updated_message = schemas.parse_message(message_body)
                 updated_message.retry_count = retry_count
+                queue.delete_messages(Entries=[self.delete_message_entry(message)])
```
skyemeedan, line 147: this isn't including the idx (which is mapped to the id) is it needed?

skyemeedan (commented):
lol, this makes sense. Does it work locally?

lib/queue/worker.py
DGaffney: yep - idx is only needed as some unique identifier in a list of items - and on the other side of this, it just defaults to 0 if not passed along, so this works fine!

DGaffney:
> lol, this makes sense. Does it work locally?

lol, yep - and yes I did test this on the QA box directly to make sure that message deletion would work, as I was figuring out what the error was in this latest deploy

skyemeedan (approved):

PR #97, opened 2024-07-10
ClassyCat in Presto: CV2 4789

ashkankzme:
## Description
This PR is for implementing the three main classycat endpoints into presto: schema_create, schema_lookup, and classify

Reference: TICKET-ID (to provide additional context)
https://meedan.atlassian.net/jira/software/c/projects/CV2/issues/CV2-4789

## How has this been tested?
Has it been tested locally? Yes
Are there automated tests? Yes, unit tests under test_classycat.py

- openai dependency added to requirements.txt
- new env vars added to .env_file
- new SQS queues needed for serving classycat requests

N/A


test/lib/model/test_classycat.py
```diff
@@ -0,0 +1,708 @@
...
+                    "event_type": "classify",
+                    "schema_id": "4a026b82-4a16-440d-aed7-bec07af12205",
+                    "items": [
+                        {
+                            "id": "11",
+                            "text": "modi and bjp want to rule india by dividing people against each other"
+                        }
+                    ]
+                },
+                "callback_url": "http://host.docker.internal:9888"
```
computermacgyver, line 699: It looks like tests are failing at an earlier stage right now around getting `CLASSYCAT_BATCH_SIZE_LIMIT`, but I suspect the tests will also fail with a timeout from `http://host.docker.internal:9888` on Linux. This has been my experience. 

> If you are using [Docker-for-mac](https://docs.docker.com/docker-for-mac/networking/#there-is-no-docker0-bridge-on-macos#i-want-to-connect-from-a-container-to-a-service-on-the-host) or [Docker-for-Windows](https://docs.docker.com/docker-for-windows/networking/#there-is-no-docker0-bridge-on-windows#i-want-to-connect-from-a-container-to-a-service-on-the-host) 18.03+, connect to your mysql service using the host host.docker.internal (instead of the 127.0.0.1 in your connection string).
>
> If you are using Docker-for-Linux 20.10.0+, you can also use the host host.docker.internal if you started your Docker container with the --add-host host.docker.internal:host-gateway option, or added the following snippet in your docker-compose.yml file :
```
extra_hosts:
    - "host.docker.internal:host-gateway"
```
More at  https://stackoverflow.com/a/24326540

lib/model/classycat_schema_create.py
```diff
@@ -0,0 +1,170 @@
...
+
+        schema["prompt"] = self.generate_prompt_from_schema(schema)
+
+        upload_file_to_s3(self.output_bucket, schema_filename, json.dumps(schema))
+        # s3_client.put_object(
+        #     Bucket=output_bucket, Key=schema_filename, Body=json.dumps(schema)
+        # )
+
+        upload_file_to_s3(self.output_bucket, schema_id_mapping_filename, json.dumps({"schema_id": schema_id}))
+        # s3_client.put_object(
```
skyemeedan, line 49: This logs that it uploaded to S3, but is commented?  Maybe could be removed if code is using upload_file_to_s3_using_filename() instead?

lib/model/classycat_schema_create.py
```diff
@@ -0,0 +1,170 @@
...
+
+        return self.base_prompt.format(
+            categories="\n".join(prompt_categories),
+            languages=languages,
+            schema_name=schema["schema_name"],
+        )
+
+
+    def process(self, message: Message) -> ClassyCatSchemaResponse:
+        # unpack parameters for create_schema
```
skyemeedan, line 107: having an example of the schema response in comments here would be really helpful

lib/schemas.py
```diff
@@ -32,6 +41,14 @@ def parse_message(message_data: Dict) -> Message:
...
+        event_type = body_data['parameters']['event_type']
+        if event_type == 'classify':
+            result_instance = ClassyCatBatchClassificationResponse(**result_data)
+        elif event_type == 'schema_lookup' or event_type == 'schema_create':
+            result_instance = ClassyCatSchemaResponse(**result_data)
+        else:
+            result_instance = ClassyCatResponse(**result_data)
     elif 'video' in model_name:
         result_instance = VideoResponse(**result_data)
     else:
```
skyemeedan, line 54: I think this else should be `elif` should raise error if unknown model name  (otherwise will silently fail on typo?)

lib/model/classycat_classify.py
```diff
@@ -0,0 +1,129 @@
...
+
+    def classify(self, prompt_template, items, max_tokens_per_item=200):
+        task_prompt = prompt_template.replace('<INSERT_ITEMS_HERE>', self.format_input_for_classification_prompt(items))
+        raw_results = self.llm_client.classify(task_prompt, len(items), max_tokens_per_item)
+        logger.info(f"Raw completion: {raw_results}")
+        return self.parse_classification_results(raw_results)
+
+    def classify_and_store_results(self, schema_id, items):
+        # load prompt from schema
+        schema = load_file_from_s3(self.output_bucket, f"{schema_id}.json")
```
skyemeedan, line 80: Idea: (maybe not relevant if s3 load is small percent of time compared to LLM roundtrip) Since we don't have a huge number of schemas, and not huge, could load them up the first time we see them and not reload from s3 for every classification event?

lib/model/classycat_classify.py
```diff
@@ -0,0 +1,129 @@
...
+
+        item_texts = [item['text'] for item in items]
+        classification_results = self.classify(prompt, item_texts)
+
+        if (classification_results is None or len(classification_results) == 0
+                or len(classification_results) != len(items)):
+            logger.info(f"Classification results: {classification_results}")
+            raise Exception(f"Not all items were classified successfully: "
+                            f"input length {len(items)}, output length {len(classification_results)}")
+
```
skyemeedan, line 93:
```suggestion
# TODO: validate response label against schema https://meedan.atlassian.net/browse/CV2-4801
```

lib/model/classycat_classify.py
```diff
@@ -0,0 +1,129 @@
...
+        final_results = [{'id': items[i]['id'], 'text': items[i]['text'], 'labels': classification_results[i]}
+                         for i in range(len(items))]
+        results_file_id = str(uuid.uuid4())
+        upload_file_to_s3(self.output_bucket, f"{schema_id}/{results_file_id}.json", json.dumps(final_results))
```
skyemeedan, line 97: Maybe the file upload should only be enabled in Live?

lib/model/classycat_classify.py
```diff
@@ -0,0 +1,129 @@
...
+        client = self.get_client()
+        completion = client.chat.completions.create(
+            model=self.model_name,
+            messages=[
+                {"role": "user", "content": task_prompt}
+            ],
+            max_tokens=(max_tokens_per_item * items_count) + 15,
+            temperature=0.5
+        )
+
```
skyemeedan, line 45:
```suggestion
# TODO: record metric here with model name and number of items submitted? (do we have a ticket for metrics?)
```

skyemeedan, line 45: Can we check here for some of the OpenRouter specific errors (like out of tokens) so the model can error instead of passing downstream?

skyemeedan (commented):
Looks great, awesome how compact it is!
* I'm still not clear how a consumer of the model will get the schemas uploaded and deployed (in the tests it is assuming that consumer has a local copy of the schema text to submit before doing classification)
* does it need some docker-compose.py entries to be able to spin up model locally?
* I think it needs more examples and validation of input and output data structures to document what needs to be submitted

test/lib/model/test_classycat.py
ashkankzme: thanks for flagging Scott. 

I think the tests failing bc of `CLASSYCAT_BATCH_SIZE_LIMIT` (or other en vars) is due to the fact that these new environment variables are not being correctly set in the test cases. I will look into what the issue is and try to figure it out.

re: callbacks not working -- I think that makes sense and is expected. the address mentioned there is a local server I have spun up to catch the callbacks from Presto. I did not commit the server with the PR code (but happy to do so if helpful), and therefore it doesn't exist to receive the results. 

I'm now realizing that the other tests have been mocking the callback functionality, and I have not done so for classycat. I will do that soon and update the PR. it should fix the timeout error.

.env_file.example
```diff
@@ -13,4 +13,9 @@ OTEL_EXPORTER_OTLP_ENDPOINT="https://api.honeycomb.io"
 OTEL_EXPORTER_OTLP_HEADERS="x-honeycomb-team=XXX"
 HONEYCOMB_API_ENDPOINT="https://api.honeycomb.io"
 REDIS_URL="redis://redis:6379/0"
-CACHE_DEFAULT_TTL=86400
\ No newline at end of file
+CACHE_DEFAULT_TTL=86400
+
+CLASSYCAT_OUTPUT_BUCKET="litterbox-dev"
```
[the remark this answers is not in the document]
ashkankzme: haha, sure we can change the bucket naming

lib/model/classycat_schema_create.py
ashkankzme: it's just old code that I have commented out and replaced with `upload_file_to_s3()`, I will remove it

lib/schemas.py
ashkankzme: I don't think it will fail silently as is, since we check for the event type when processing the request later. please look at `classycat.py/process()`. 

I did this so I can respond with the same error structure as other classycat request, otherwise you may not receive an error directly from presto or it may be in a different format.

lib/model/classycat_classify.py
ashkankzme: a good idea but probably premature optimization at this point. we should def consider caching schemas if we ever need performance improvements here.

test/lib/model/test_classycat.py
DGaffney: Yes, my personal preference as well is mocking callback functionality for intra-container tests

lib/model/classycat_classify.py
ashkankzme: I would like to be able to test that locally as well and make sure the results are being properly stored in minio

lib/s3.py
```diff
@@ -3,26 +3,21 @@
...
+    s3_client = get_s3_client()
     # Extract the file name from the local file path
     try:
         s3_client.head_bucket(Bucket=bucket)
     except ClientError as e:
         if e.response['Error']['Code'] == 'NoSuchBucket' or int(e.response['Error']['Code']) in [403, 404]:
             # Create the bucket since it does not exist
             s3_client.create_bucket(Bucket=bucket)
-            logger.info(f'Created bucket {bucket} in MinIO.')
+            logger.info(f'Created bucket {bucket} in MinIO.') # is this accurate? i.e. is this code only for minio?
```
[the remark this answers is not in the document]
DGaffney: I believe the default pattern we established was to create the bucket locally but not in deployed environments, and that the create_bucket would fail because of permissions errors. Tagging @sonoransun to confirm/deny

.env_file.example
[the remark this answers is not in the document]
DGaffney: +1 for keeping it litterbox, lol

lib/model/classycat_classify.py
DGaffney: From my quick read, that seems like we'd need maybe some global memoized store of these that persists across process iterations? If so that's a fairly substantive departure from previous assumptions about statelessness?

lib/model/classycat_classify.py
DGaffney: I think we'd still need to upload in Minio to be fully functional? Is that acceptable or no?

lib/model/classycat_classify.py
ashkankzme: trying to better understand the scope, can you please explain a bit more which cases this would cover, and what the end goal would be?

skyemeedan: Currently timpani sees errors like 'not enough credits', but doesn't know what that means, so it is going to keep sending additional requests.  I think if classcat is unable to do any processing, it should be reporting that as an exception, and perhaps crashing/going offline to signal for help?

lib/model/classycat_classify.py
skyemeedan: I think it is fine to leave it enabled in any of the environments.  But i was thinking it would probably only be the data in live that we'd want to be saving for possible model training?

lib/schemas.py
skyemeedan: oh I was wrong. I understand now this is just mesage validation, I thought it was parsing incoming messages.  I lost a couple of hours debugging because I was sending messages with the model 'yake' instead of 'yake_keywords'  and it just silently accepted ... but this isn't the place to enforce that.

DGaffney: hm, for what its worth @skyemeedan we *could* raise an error if the model name is unknown to this function...

.env_file.example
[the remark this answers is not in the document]
skyemeedan: classycatlitter-qa?

lib/model/classycat_classify.py
ashkankzme: I see. I agree that it is helpful for the consumer of this service to distinguish between when it is ok to retry upon errors vs when it's a matter of service downtime and it's not ok to retry. I'll try to incorporate this into the code!

lib/s3.py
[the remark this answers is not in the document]
ashkankzme: I just updated the log message to be a bit clearer (feel free to push back) -- it now says creating a bucket in s3 (or minio, if running locally)

lib/schemas.py
ashkankzme: @DGaffney we do respond with an error to the caller in this case, it's just handled in the process function instead of here. happy to discuss further!

lib/model/classycat_classify.py
ashkankzme: totally fair! maybe the best solution would be to cache the schemas, that way we won't have to worry about preserving state or not. but I still think this will not be a bottleneck anytime soon, since per Skye's comments, this is not a substantial part of the request lifecycle (it takes 40+ seconds to hear back from LLMs) and therefore optimizing it would not bring meaningful improvements to the endpoint at this point.

lib/model/classycat_classify.py
ashkankzme: I'm gonna leave it as is for complete testing of functionality in local development

lib/model/classycat_classify.py
computermacgyver: Let's defer this until the PR for consistent error/status output and error handling is in place (this is in separate work)

lib/model/classycat_classify.py
computermacgyver: Sounds right. Let's not try to optimize the schema loading right now.

lib/schemas.py
computermacgyver: Just to clarify, @ashkankzme , I believe Devin was referring to a Presto model (YAKE, Video, Audio, etc.) and I believe you're referring to an LLM model. I think that's a good idea to respond with an error if the Presto model is unknown. That, however, can be a separate ticket / PR

computermacgyver (approved):
Looks great. Let's get this merged and on QA. I want to make sure we do a full test of video similarity on QA given this touches S3

PR #99, opened 2024-07-24
CV2-4978: Local callback server for receiving presto callback responses

ashkankzme:
## Description
Simple python code for a callback server that spins up on an input port (9888 by default) and receives and prints post requests, which can be used as the callback server for testing/debugging presto locally.

Reference: TICKET-ID (to provide additional context)
[CV2-4978](https://meedan.atlassian.net/browse/CV2-4978)

## How has this been tested?
This has been tested while developing classycat in presto and works fine

No

Only use this locally, not to be used in production or security sensitive use cases

[CV2-4978]: https://meedan.atlassian.net/browse/CV2-4978?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/extra/callback_server.py
+++ b/extra/callback_server.py
@@ -1,0 +1,30 @@
+from http.server import HTTPServer, BaseHTTPRequestHandler
+import sys
+
+
+class RequestHandler(BaseHTTPRequestHandler):
+    def do_POST(self):
+        content_length = int(self.headers['Content-Length'])
+        post_data = self.rfile.read(content_length)
+
+        print("Received POST request:")
+        print(post_data.decode('utf-8'))
+
+        self.send_response(200)
+        self.send_header('Content-type', 'text/plain')
+        self.end_headers()
+        self.wfile.write(b"{\"message\": \"Message Received\"}")
+
+
+def run_server(port=9888):
+    server_address = ('', port)
+    httpd = HTTPServer(server_address, RequestHandler)
+    print(f"Server running on port {port}")
+    httpd.serve_forever()
+
+
+if __name__ == '__main__':
+    port = 9888
+    if len(sys.argv) > 1:
+        port = int(sys.argv[1])
+    run_server(port)
--- a/extra/callback_server.py
+++ b/extra/callback_server.py
@@ -1,7 +1,7 @@
 from http.server import HTTPServer, BaseHTTPRequestHandler
 import sys, os
 
-DEFAULT_PORT = os.getenv('LOCAL_CALLBACK_PORT', 9888)
+DEFAULT_PORT = int(os.getenv('LOCAL_CALLBACK_PORT', 9888))
 
 
 class RequestHandler(BaseHTTPRequestHandler):
--- a/extra/callback_server.py
+++ b/extra/callback_server.py
@@ -1,5 +1,7 @@
 from http.server import HTTPServer, BaseHTTPRequestHandler
-import sys
+import sys, os
+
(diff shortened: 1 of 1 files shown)


computermacgyver (approved):
Do we need the `__init__.py` file? I think we can remove that as we're not running this as a Python module.

extra/callback_server.py
```diff
@@ -0,0 +1,30 @@
+from http.server import HTTPServer, BaseHTTPRequestHandler
+import sys
+
+
+class RequestHandler(BaseHTTPRequestHandler):
+    def do_POST(self):
```
DGaffney, line 6: rename the function to something a bit clearer

extra/callback_server.py
```diff
@@ -0,0 +1,30 @@
...
+        print("Received POST request:")
+        print(post_data.decode('utf-8'))
+
+        self.send_response(200)
+        self.send_header('Content-type', 'text/plain')
+        self.end_headers()
+        self.wfile.write(b"{\"message\": \"Message Received\"}")
+
+
+def run_server(port=9888):
```
DGaffney, line 19: The port needs to be a parameter exposed as an env variable, also if you have 9888 written out twice in this file, better to just put it as a constant at the top. Look at this example from `cache.py` as a "canonical" way to set this type of env var: `DEFAULT_TTL = int(get_environment_setting("CACHE_DEFAULT_TTL") or 24*60*60)`

extra/callback_server.py
```diff
@@ -0,0 +1,30 @@
+from http.server import HTTPServer, BaseHTTPRequestHandler
+import sys
+
+
+class RequestHandler(BaseHTTPRequestHandler):
+    def do_POST(self):
+        content_length = int(self.headers['Content-Length'])
+        post_data = self.rfile.read(content_length)
+
+        print("Received POST request:")
```
DGaffney, line 10: Instead of print, maybe an app logger of some sort?

DGaffney (changes requested):

extra/callback_server.py
ashkankzme: I'm afraid that's not possible, this is the function we override to handle post requests, as defined in native python docs for this library. more info here: https://docs.python.org/3/library/http.server.html

extra/callback_server.py
ashkankzme: since we only want to run this locally, I prefer to keep it as a light weight independent script for running local applications without the need to install/import any custom/additional libs. I don't think the print instead of logger will cause any issues?

extra/callback_server.py
ashkankzme: makes sense, just added the environment variable with a default value of 9888 as suggested
I'd like to keep the cmd argument option open, which means we first look at env var, default to 9888 if not set, and can change that either through setting the env var or passing it as an arg to the server when running it in the command line.

extra/callback_server.py
DGaffney: Thanks I hate it lol

extra/callback_server.py
DGaffney: That's fine for now - if we add anything else I think adding it as a logger would be appropriate - it's a standard library so wouldn't add any more dependencies, FWIW

DGaffney (approved):

PR #101, opened 2024-07-26
double check returned categories to make sure they are in-schema + tests

ashkankzme:
## Description
Code and tests for verifying the returned labels belong to the same set as defined in the classycat schema.
In case some items in a batch have out-of-schema labels, those items will be omitted from the results, however the rest of items that were classified as expected will be returned. In case all the generated labels are out of schema, an error will be returned, marking the operation as failure.

Reference: [CV2-4801](https://meedan.atlassian.net/jira/software/c/projects/CV2/issues/CV2-4801)

## How has this been tested?
Has been tested locally and through unit tests.
However the condition where LLM returns out of schema labels is difficult to control and reproduce.
But the scenario has been reproduced in unit tests and it's verified the code functions as expected in case of out of schema labels.

This code requires classycat functionality in presto to be deployed and builds on top of the classycat branch

No concerns

[CV2-4801]: https://meedan.atlassian.net/browse/CV2-4801?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/lib/model/classycat_classify.py
+++ b/lib/model/classycat_classify.py
@@ -130,6 +130,11 @@ class Model(Model):
         # and items with no labels can be retried later by the user, indicated by an empty list for labels
         permitted_labels = [topic['topic'] for topic in schema['topics']] + ['Other', 'Unsure']
         for result in final_results:
+
+            # log the items that had at least one out-of-schema label
+            if not all([label in permitted_labels for label in result['labels']]):
+                logger.error(f"Item {result['id']} had out-of-schema labels: {result['labels']}, permitted labels: {permitted_labels}")
+
             result['labels'] = [label for label in result['labels'] if label in permitted_labels]
 
         if not all([len(result['labels']) == 0 for result in final_results]):
--- a/lib/model/classycat_classify.py
+++ b/lib/model/classycat_classify.py
@@ -121,9 +121,18 @@ class Model(Model):
             logger.info(f"Classification results: {classification_results}")
             raise Exception(f"Not all items were classified successfully: "
                             f"input length {len(items)}, output length {len(classification_results)}")
-        # TODO: validate response label against schema https://meedan.atlassian.net/browse/CV2-4801
+
         final_results = [{'id': items[i]['id'], 'text': items[i]['text'], 'labels': classification_results[i]}
                          for i in range(len(items))]
+
+        # filtering out the results that have out-of-schema labels
+        permitted_labels = [topic['topic'] for topic in schema['topics']] + ['Other', 'Unsure']
+        final_results = [result for result in final_results if all(label in permitted_labels for label in result['labels'])]
+
+        if len(final_results) == 0:
+            logger.info(f"The returned classifications did not produce labels from the schema: {items}")
+            raise Exception(f"No items were classified successfully")
+
         results_file_id = str(uuid.uuid4())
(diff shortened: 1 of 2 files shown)


computermacgyver (approved):

lib/model/classycat_classify.py
```diff
@@ -121,9 +121,18 @@ def classify_and_store_results(self, schema_id, items):
...
         final_results = [{'id': items[i]['id'], 'text': items[i]['text'], 'labels': classification_results[i]}
                          for i in range(len(items))]
+
+        # filtering out the results that have out-of-schema labels
+        permitted_labels = [topic['topic'] for topic in schema['topics']] + ['Other', 'Unsure']
+        final_results = [result for result in final_results if all(label in permitted_labels for label in result['labels'])]
+
+        if len(final_results) == 0:
+            logger.info(f"The returned classifications did not produce labels from the schema: {items}")
+            raise Exception(f"No items were classified successfully")
```
DGaffney, line 134: Seems like this will percolate upwards and eventually return an error response, yes? If so I think it's fine!

DGaffney (approved):

lib/model/classycat_classify.py
```diff
@@ -121,9 +121,18 @@ def classify_and_store_results(self, schema_id, items):
...
             raise Exception(f"Not all items were classified successfully: "
                             f"input length {len(items)}, output length {len(classification_results)}")
-        # TODO: validate response label against schema https://meedan.atlassian.net/browse/CV2-4801
+
         final_results = [{'id': items[i]['id'], 'text': items[i]['text'], 'labels': classification_results[i]}
                          for i in range(len(items))]
+
+        # filtering out the results that have out-of-schema labels
+        permitted_labels = [topic['topic'] for topic in schema['topics']] + ['Other', 'Unsure']
+        final_results = [result for result in final_results if all(label in permitted_labels for label in result['labels'])]
```
skyemeedan, line 130: maybe helpful to log a warning here for the ones that got dropped?

lib/model/classycat_classify.py
```diff
@@ -121,9 +121,18 @@ def classify_and_store_results(self, schema_id, items):
...
+
+        if len(final_results) == 0:
+            logger.info(f"The returned classifications did not produce labels from the schema: {items}")
```
skyemeedan, line 133: I think this should be logger.error?

skyemeedan (approved):

lib/model/classycat_classify.py
ashkankzme: we decided to not drop any of the items, and instead return an empty list for labels, for items that are not labeled properly. it would be good to track percentage of failed requests/classifications but I would count that as the metrics collection work.

lib/model/classycat_classify.py
ashkankzme: this no longer exists since code changed a bit

lib/model/classycat_classify.py
ashkankzme: this code will be updated once I merge the error handling PR (in progress)

lib/model/classycat_classify.py
skyemeedan: I agree returning empty list right approach for now, but I think it is important to not do it silently, so I think just logging a warning each time should be easy and sufficient.  Maybe it is rare enough we won't even need metric

PR #103, opened 2024-08-05
CV2-5011 refactors for making alegre dual purpose on text encoding

DGaffney:
## Description
Minor tweak identified when running text encodings.

Reference: CV2-5011

## How has this been tested?
Tested locally and confirmed to work

Changes no dependencies

Does not alter security posture.

### DIFF
--- a/lib/model/generic_transformer.py
+++ b/lib/model/generic_transformer.py
@@ -33,4 +33,4 @@ class GenericTransformerModel(Model):
         """
         Vectorize the text! Run as batch.
         """
-        return {"hash_value": self.model.encode(texts).tolist()}
+        return self.model.encode(texts).tolist()
--- a/lib/model/generic_transformer.py
+++ b/lib/model/generic_transformer.py
@@ -1,9 +1,10 @@
 import os
-from typing import Union, Dict, List
+from typing import Union, List
 from sentence_transformers import SentenceTransformer
 from lib.logger import logger
 from lib.model.model import Model
 from lib import schemas
+from lib.cache import Cache
 
 class GenericTransformerModel(Model):
     def __init__(self, model_name: str):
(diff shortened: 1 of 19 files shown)


caiosba (commented):
Sorry, I'm not familiarized with this code - but, does this change in the return require any change on Alegre side?

DGaffney:
> Sorry, I'm not familiarized with this code - but, does this change in the return require any change on Alegre side?

It does not - the only thing that's happening here is basically updating vectorizers to return what is now the current specification for how responses from fingerprinters should look - this one got out of sync since we hadn't been using it for anything. Mostly just tagging you here for visibility so you can be aware of the magnitude of changes across each repo, don't worry too much about the specifics in this repo!

lib/model/generic_transformer.py
ashkankzme, line 1: I realize this file has existed for a while, but is there a reason we override the respond() function instead of process()?
this would make this model unique and different from other Presto endpoints, and disables cache and error handling bc the standard get_response() function won't be called from model.py.

But I may be missing something?

DGaffney: Yes - the reason we override is because we want to, from the get-go, natively be able to process jobs in batch on transformers instead of just walking through items one-by-one like we do with the others. I'll look into a refactor to keep our caching and error handling working for text though - good catch

skyemeedan (approved):
> the only thing that's happening here is basically updating vectorizers to return what is now the current specification for how responses from fingerprinters should look - this one got out of sync since we hadn't been using it for anything. 

I'm not  very clear how the output changed, but it looks like it will handle caching better? Is the current specification somewhere I can check if it matches?   I'm just assuming it works because you said you tested it locally

DGaffney:
> > the only thing that's happening here is basically updating vectorizers to return what is now the current specification for how responses from fingerprinters should look - this one got out of sync since we hadn't been using it for anything.
> 
> I'm not very clear how the output changed, but it looks like it will handle caching better? Is the current specification somewhere I can check if it matches? I'm just assuming it works because you said you tested it locally

That's basically it @skyemeedan - the function calling the fingerprinting function expects a slightly different response. Once we have the fixes from @ashkankzme on the PR he's working on it should be a bit easier to highlight conceptual float like this case.

computermacgyver (approved):
This looks fine for now. Thanks!
When we work on the bulk endpoints for Alegre, I think we're going to want one request to Presto to have a list of multiple text items to be vectorized. All of the vectors for those items should then be returned in one callback. I think some of the changes here may complicate that, but we can address it when we get to the bulk endpoints on Alegre.

test/lib/queue/test_queue.py
```diff
@@ -31,8 +31,8 @@ class TestQueueWorker(unittest.TestCase):
     @patch('lib.helpers.get_environment_setting', return_value='us-west-1')
     @patch('lib.telemetry.OpenTelemetryExporter.log_execution_time')
     def setUp(self, mock_log_execution_time, mock_get_env_setting, mock_boto_resource):
-        self.model = GenericTransformerModel(None)
-        self.model.model_name = "generic"
+        self.model = AudioModel()
+        self.model.model_name = "audio"
```
ashkankzme, line 35: can we make sure model_name follows the same format as everywhere else? these names would break the tests in the new refactor

PR #104, opened 2024-08-05
CV2-4952: ClassyCat Error Handling and Metrics

ashkankzme:
## Description
Error handling for ClassyCat using new error handling / exception mechanisms in Presto
ClassyCat modules now raise unhandled "PrestoBaseException" objects that include an http error code to properly respond to consumer requests.

Reference: [CV2-4952](https://meedan.atlassian.net/jira/software/c/projects/CV2/issues/CV2-4952)

## How has this been tested?
Has been tested locally and in unit tests

No

We return stack traces included in error messages, which raises security concerns.

[CV2-4952]: https://meedan.atlassian.net/browse/CV2-4952?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/lib/model/classycat_classify.py
+++ b/lib/model/classycat_classify.py
@@ -8,6 +8,7 @@ from lib.logger import logger
 from lib.model.model import Model
 from lib.schemas import Message, ClassyCatBatchClassificationResponse
 from lib.s3 import load_file_from_s3, file_exists_in_s3, upload_file_to_s3
+from lib.base_exception import PrestoBaseException
 
 
 class LLMClient:
@@ -84,7 +85,7 @@ class Model(Model):
         elif client_type == 'openrouter':
             return OpenRouterClient(model_name)
         else:
-            raise Exception(f"Unknown client type: {client_type}")
+            raise PrestoBaseException(f"Unknown LLM client type {client_type}", 500)
 
     def format_input_for_classification_prompt(self, items):
         return '\n'.join([f"<ITEM_{i}>{item}</ITEM_{i}>" for i, item in enumerate(items)])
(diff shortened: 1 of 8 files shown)


lib/model/classycat_classify.py
```diff
@@ -84,7 +87,7 @@ def get_llm_client(self, client_type, model_name):
         elif client_type == 'openrouter':
             return OpenRouterClient(model_name)
         else:
-            raise Exception(f"Unknown client type: {client_type}")
+            raise PrestoBaseException(f"Unknown LLM client type {client_type}", 500)
```
skyemeedan, line 90: should this be a 400 code of some type?

lib/model/model.py
```diff
@@ -64,7 +68,10 @@ def get_response(self, message: schemas.Message) -> schemas.GenericItem:
                 result = self.process(message)
                 Cache.set_cached_result(message.body.content_hash, result)
             except Exception as e:
-                return self.handle_fingerprinting_error(e)
+                if isinstance(e, PrestoBaseException):
```
skyemeedan, line 71: since we are (I think) nicely bubbling up errors to this level,  should logger.exception() here so that it will still get logged to Sentry from the presto side?

test/lib/model/test_classycat.py
```diff
@@ -562,9 +563,17 @@ def test_classify_fail_wrong_response_format(self, file_exists_in_s3_mock, uploa
             }
         }
         classify_message = schemas.parse_message(classify_input)
-        result = self.classycat_model.process(classify_message)
 
-        self.assertEqual(result.responseMessage, "Error classifying items: list index out of range")
+        try:
+            result = self.classycat_model.process(classify_message)
```
skyemeedan, line 568: side comment:  there is an assertRaises pattern in the unitTest framework that can be helpful for these

skyemeedan (commented):
This looks awesome to me! I do think we should still logger.exception() when handling the errors at the top level

lib/model/model.py
ashkankzme: I think this exists on line 58, no?

test/lib/model/test_classycat.py
ashkankzme: great idea, updated

lib/model/model.py
skyemeedan: oh, I didn't realize that is what capture_custom_messages does.  In Presto we are explicitly capturing message so it will go to Sentry only and not normal python logs?  If so maybe that capture message argument should be 'error' instead of 'info'

ashkankzme: done! (changed info to error)

overall, I also think it's better to log errors when they happen. makes debugging easier, and we also won't lose the line number on which the error logging happens. for all of classycat errors, I made sure we log the errors where they happen.

DGaffney (approved):
:shipit:

PR #105, opened 2024-08-08
Cv2 4953

ashkankzme:
## Description
Store successful classification results of ClassyCat in Alegre

Reference: [CV2-4953](https://meedan.atlassian.net/jira/software/c/projects/CV2/issues/CV2-4953)

## How has this been tested?
Has been tested locally and in unit tests.

New SSM parameters need to be added to Presto for the Alegre endpoint URL

None

### DIFF
--- a/lib/model/classycat_classify.py
+++ b/lib/model/classycat_classify.py
@@ -140,24 +140,27 @@ class Model(Model):
 
             result['labels'] = [label for label in result['labels'] if label in permitted_labels]
 
+        # if there is at least one item with labels, save the results to s3
         if not all([len(result['labels']) == 0 for result in final_results]):
             results_file_id = str(uuid.uuid4())
             upload_file_to_s3(self.output_bucket, f"{schema_id}/{results_file_id}.json", json.dumps(final_results))
 
-            # save content and context
-            # content is text, doc_id is unique id, and context is input id, labels, schema_id, and model name
-            final_results = {'documents': [
+            # prepare the final results to be stored in alegre
+            # save "content" and "context"
+            # content is text, doc_id is the item's unique id, and context is input id, labels, schema_id, and model name
+            final_results_to_be_stored_in_alegre = {'documents': [
                 {'doc_id': str(uuid.uuid4()),  # adding a unique id for each item to not rely on the input id for uniqueness
                  'content': items[i]['text'],
                  'context': {
                      'input_id': items[i]['id'],
-                     'labels': classification_results[i],
+                     'labels': final_results[i]['labels'],
                      'schema_id': schema_id,
                      'model_name': self.llm_client.model_name}}
                 for i in range(len(items))]}
 
-            # call alegre endpoint to store the results: /text/bulk_similarity
-            httpx.post('http://alegre:3100/text/bulk_similarity', json=final_results)  # todo fix endpoint and headers
+            # call alegre endpoint to store the results: /text/bulk_similarity/
+            alegre_url = os.getenv('ALEGRE_URL')
+            httpx.post(alegre_url + '/text/bulk_similarity/', json=final_results_to_be_stored_in_alegre)
 
         return final_results
 
--- a/lib/model/classycat_classify.py
+++ b/lib/model/classycat_classify.py
@@ -144,20 +144,20 @@ class Model(Model):
             results_file_id = str(uuid.uuid4())
             upload_file_to_s3(self.output_bucket, f"{schema_id}/{results_file_id}.json", json.dumps(final_results))
 
-        # save content and context
-        # content is text, doc_id is unique id, and context is input id, labels, schema_id, and model name
-        final_results = {'documents': [
-            {'doc_id': str(uuid.uuid4()),  # adding a unique id for each item to not rely on the input id for uniqueness
-             'content': items[i]['text'],
-             'context': {
-                 'input_id': items[i]['id'],
-                 'labels': classification_results[i],
-                 'schema_id': schema_id,
-                 'model_name': self.llm_client.model_name}}
-            for i in range(len(items))]}
+            # save content and context
+            # content is text, doc_id is unique id, and context is input id, labels, schema_id, and model name
+            final_results = {'documents': [
+                {'doc_id': str(uuid.uuid4()),  # adding a unique id for each item to not rely on the input id for uniqueness
+                 'content': items[i]['text'],
+                 'context': {
+                     'input_id': items[i]['id'],
+                     'labels': classification_results[i],
+                     'schema_id': schema_id,
+                     'model_name': self.llm_client.model_name}}
+                for i in range(len(items))]}
 
-        # call alegre endpoint to store the results: /text/bulk_similarity
-        httpx.post('http://alegre:9888/text/bulk_similarity', json=final_results) # todo fix endpoint and headers
+            # call alegre endpoint to store the results: /text/bulk_similarity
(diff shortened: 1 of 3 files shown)


lib/model/classycat_classify.py
```diff
@@ -137,17 +140,33 @@ def classify_and_store_results(self, schema_id, items):
...
+                 'context': {
+                     'input_id': items[i]['id'],
+                     'labels': final_results[i]['labels'],
+                     'schema_id': schema_id,
+                     'model_name': self.llm_client.model_name}}
+                for i in range(len(items))]}
+
+            # call alegre endpoint to store the results: /text/bulk_similarity/
+            alegre_url = os.getenv('ALEGRE_URL')
+            httpx.post(alegre_url + '/text/bulk_similarity/', json=final_results_to_be_stored_in_alegre)
```
skyemeedan, line 163: ooh, does this insert endpoint work?  maybe I can switch timpani to use it as well when we convert to batch mode!

lib/model/classycat_classify.py
```diff
@@ -137,17 +140,33 @@ def classify_and_store_results(self, schema_id, items):
...
+            # save "content" and "context"
+            # content is text, doc_id is the item's unique id, and context is input id, labels, schema_id, and model name
+            final_results_to_be_stored_in_alegre = {'documents': [
+                {'doc_id': str(uuid.uuid4()),  # adding a unique id for each item to not rely on the input id for uniqueness
+                 'content': items[i]['text'],
+                 'context': {
+                     'input_id': items[i]['id'],
+                     'labels': final_results[i]['labels'],
+                     'schema_id': schema_id,
+                     'model_name': self.llm_client.model_name}}
```
skyemeedan, line 158: I think this needs to be the model name for the text vectorization? i.e. `"paraphrase-multilingual-mpnet-base-v2"`?
These are the parameters timpani used:
https://github.com/meedan/timpani/blob/6021d4fcb251d83ae48ed2c1566a16fad6971450/timpani/model_service/alegre_wrapper_service.py#L122

skyemeedan (commented):
Did this work when running locally with Alegre (i.e. a manual integration test)?
* I think it needs the key to specify which model will be used for vectorization

Comment:
I'm still not entirely sure why we are storing again in Alegre? I believe the text, the associated vector, the cluster, and the labels are already stored in alegre or calling systems by the time we call classycat.  Have we ruled out appending an additional 'context' to existing items to store the labels?  Or having a separate "label propagation" function in classycat that accepts  an item, and a list of items with associated labels (assumed to be in a cluster) returns the propagated label or a new one if none available?

lib/model/classycat_classify.py
computermacgyver: Within `context`, `model_name` (or any parameter) can be anything. We will, however, need to specify the `models` parameter for Alegre itself to know how to do vectorization. That parameter is currently missing.

lib/model/classycat_classify.py
```diff
@@ -137,17 +140,33 @@ def classify_and_store_results(self, schema_id, items):
...
             results_file_id = str(uuid.uuid4())
             upload_file_to_s3(self.output_bucket, f"{schema_id}/{results_file_id}.json", json.dumps(final_results))
 
-        return final_results
+            # prepare the final results to be stored in alegre
+            # save "content" and "context"
+            # content is text, doc_id is the item's unique id, and context is input id, labels, schema_id, and model name
+            final_results_to_be_stored_in_alegre = {'documents': [
+                {'doc_id': str(uuid.uuid4()),  # adding a unique id for each item to not rely on the input id for uniqueness
+                 'content': items[i]['text'],
```
computermacgyver, line 153: Annoyingly, this parameter needs to be `text`, I believe.

lib/model/classycat_classify.py
computermacgyver: We think so, but @DGaffney is currently moving all vectorization to Presto and will need to ensure this endpoint continues to work after that migration :sweat_smile:

computermacgyver (changes requested):

lib/model/classycat_classify.py
ashkankzme: At least for /text/bulk_similarity/ we replace `text` with `content`:

```
# Rename "text" to "content" if present
if 'text' in params:
  params['content'] = params.get('text')
  del params["text"]
```

We can use both `content` and `text` for bulk but it ultimately gets renamed to `content` in Alegre.

lib/model/classycat_classify.py
ashkankzme: good catch Scott, I did not know that we should specify model_name, and on local the documents are being index only by Elasticsearch (full text search). I will update the code and redo the tests to make sure vectorization works on local.

lib/model/classycat_classify.py
ashkankzme: makes sense Scott, I added a comment on Devin's ticket to make sure he sees it. 

@skyemeedan as of this moment the endpoint works fine locally for me, feel free to test it out!

ashkankzme:
> Did this work when running locally with Alegre (i.e. a manual integration test)?
> 
> * I think it needs the key to specify which model will be used for vectorization
> 
> Comment: I'm still not entirely sure why we are storing again in Alegre? I believe the text, the associated vector, the cluster, and the labels are already stored in alegre or calling systems by the time we call classycat. Have we ruled out appending an additional 'context' to existing items to store the labels? Or having a separate "label propagation" function in classycat that accepts an item, and a list of items with associated labels (assumed to be in a cluster) returns the propagated label or a new one if none available?

We plan to use Alegre for KNN similarity queries, since we already have Open/ElasticSearch set up for that service, makes a lot of sense. I agree that there are optimizations to be made here, but I rather not concern ClassyCat with those atm, we are still trying to achieve basic functionality in this ticket. 

Also to address the issue of already having the items stored in Alegre prior to calling classycat, I think it's a lot easier to make sure that Alegre updates existing records vs. adding duplicate ones, instead of adding additional endpoints and complexity to classycat. We can do that by allowing an optional "doc_id" for each record submitted to classycat. If that field exists in the call, I will make sure to pass that on to Alegre (as opposed to generate a new unique id) and ensure that Alegre does "upserts" rather than "inserts" into the DB. As mentioned before, I think this is work for a another time and should be a separate ticket.

lib/model/classycat_classify.py
DGaffney: I'm not sure if we want to support this after we move to presto vectorization. This is a blocking, bulk query which breaks expectations in Presto-based vectorization doubly (not-blocking, and single query per item). This is, to my knowledge, the only location in Check (minus the re-index job in Check-API) that uses `bulk_similarity`, and if we could get *away* from this pattern, that would be vastly preferable. Is there any way we can move this? Otherwise we'll be signing up for new complicated measures to support this long term.

DGaffney (changes requested):
Lets pretty pretty please not use bulk_similarity for this - I know we went through how to use it the other day but I didn't fully appreciate that this was part of new functionality that would be coming through a PR

lib/model/classycat_classify.py
DGaffney: I don't think it is reasonable to take on making async endpoints default bulk until after we have made text work on Presto. We've shunted in secondary features to this refactor before and the net effect is that they have greatly complicated our existing migration plan.

skyemeedan: If supporting bulk requires some refactoring of presto payload structures, wouldn't it be easier to do that before we are supporting the full text processing in live?

PR #107, opened 2024-08-18
adding auto langdetect and cleaning up keywords extracted by yake

ahmednasserswe:
## Description
applying the following to improve yake keywords extraction 

1- Replace ’ with ' and similar for other characters

2- Keep the longest keyword of if there is an overlap between two keywords. (“good” and “good morning” keep “good morning”) 

3- If language is not specified or is “auto” perform language detection with CLD

Reference: CV2-4909 (to provide additional context)

## How has this been tested?
locally + already existing tests

### DIFF
--- a/lib/model/yake_keywords.py
+++ b/lib/model/yake_keywords.py
@@ -7,8 +7,26 @@ from lib.model.model import Model
 from lib import schemas
 
 import yake
+from langdetect import detect
 
 class Model(Model):
+
+    def keep_largest_overlapped_keywords(self, keywords):
+        cleaned_keywords = []
+
+        for i in range(len(keywords)):
+            keep_keyword = True
+            for j in range(len(keywords)):
+                current_keyword = keywords[i][0]
+                other_keyword = keywords[j][0]
+                if len(other_keyword) > len(current_keyword):
+                    if other_keyword.find(current_keyword) >= 0:
+                        keep_keyword = False
+                        break
+            if keep_keyword:
+                cleaned_keywords.append(keywords[i])
+        return cleaned_keywords
+    
     def run_yake(self, text: str,
                  language: str,
                  max_ngram_size: int,
@@ -26,15 +44,27 @@ class Model(Model):
         :param num_of_keywords: int
         :returns: str
         """
+        ### if language is set to "auto", auto-detect it.
+        if language == 'auto':
+            language = detect(text)
+        ### replace special characters
+        text.replace("`", "'")
+        text.replace("‘", "'")
+        text.replace("“", "\"")
+        ### extract keywords
         custom_kw_extractor = yake.KeywordExtractor(lan=language, n=max_ngram_size, dedupLim=deduplication_threshold,
                                                     dedupFunc=deduplication_algo, windowsSize=window_size,
                                                     top=num_of_keywords, features=None)
-        return {"keywords": custom_kw_extractor.extract_keywords(text)}
+
+        ### Keep the longest keyword of if there is an overlap between two keywords.
+        keywords = custom_kw_extractor.extract_keywords(text)
+        keywords = self.keep_largest_overlapped_keywords(keywords)
+        return {"keywords": keywords}
 
     def get_params(self, message: schemas.Message) -> dict:
         params = {
             "text": message.body.text,
-            "language": message.body.parameters.get("language", "en"),
+            "language": message.body.parameters.get("language", "auto"),
             "max_ngram_size": message.body.parameters.get("max_ngram_size", 3),
             "deduplication_threshold": message.body.parameters.get("deduplication_threshold", 0.25),
             "deduplication_algo": message.body.parameters.get("deduplication_algo", 'seqm'),
--- a/lib/model/yake_keywords.py
+++ b/lib/model/yake_keywords.py
@@ -49,7 +49,9 @@ class Model(Model):
         ### replace special characters
         replacement = {"`": "'",
                        "‘": "'",
-                       "“": "\""}
+                       "’": "'",
+                       "“": "\"",
+                       "”": "\""}
         for k, v in replacement.items():
             text = text.replace(k, v)
         ### extract keywords
--- a/lib/model/yake_keywords.py
+++ b/lib/model/yake_keywords.py
@@ -25,14 +25,13 @@ class Model(Model):
             if keep_keyword:
                 cleaned_keywords.append(keywords[i])
         return cleaned_keywords
+
     def normalize_special_characters(self, text):
         replacement = {"`": "'",
                        "‘": "'",
                        "’": "'",
                        "“": "\"",
                        "”": "\""}
-
-
         for k, v in replacement.items():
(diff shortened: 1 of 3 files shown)


DGaffney (changes requested):
Please write unit tests for `keep_largest_overlapped_keywords`.

computermacgyver:
@ahmednasserswe Per the Jira ticket, let's use CLD - https://pypi.org/project/gcld3/ , which is a newer/better library than `langdetect`

lib/model/yake_keywords.py
```diff
@@ -7,8 +7,26 @@
 from lib import schemas
 
 import yake
+from langdetect import detect
```
computermacgyver, line 10: Use https://github.com/google/cld3 , not langdetect ?

lib/model/yake_keywords.py
```diff
@@ -7,8 +7,26 @@
...
 
 class Model(Model):
+
+    def keep_largest_overlapped_keywords(self, keywords):
```
computermacgyver, line 14: An easier to read implementation here would be to first sort keywords by length. Then the loop can only compare each keyword to the keywords that are the same or longer length.

This function should have unit tests. I don't think we'd see this situation, but if we had the same keyword appear twice in a list (e.g., `['Apple', 'Apple', 'Banana']`), I think we would not include `Apple` at all in the output as this is currently written (a unit test would confirm this). We should include one instance, however.

lib/model/yake_keywords.py
```diff
@@ -26,15 +44,27 @@ def run_yake(self, text: str,
         :param num_of_keywords: int
         :returns: str
         """
+        ### if language is set to "auto", auto-detect it.
+        if language == 'auto':
+            language = detect(text)
+        ### replace special characters
+        text.replace("`", "'")
+        text.replace("‘", "'")
+        text.replace("“", "\"")
```
computermacgyver, line 53: These lines have no effect. `str.replace` does not modify the original string. It returns a new string; so, these lines need to be `text = text.replace(...)` . Please write test code and verify it works.

We need to replace `“` and `”` . Consider having a `replacements` dictionary that can be looped over. e.g.,
```
replacement = { "‘": "'", ....}
for k,v in replacement.items():
    text=text.replace(k,v)
```

computermacgyver (changes requested):

lib/model/yake_keywords.py
```diff
@@ -26,15 +43,29 @@ def run_yake(self, text: str,
         :param num_of_keywords: int
         :returns: str
         """
+        ### if language is set to "auto", auto-detect it.
+        if language == 'auto':
+            language = cld3.get_language(text).language
+        ### replace special characters
+        replacement = {"`": "'",
+                       "‘": "'",
+                       "“": "\""}
```
computermacgyver, line 52: We're still missing one of the smart double quotes here:
> We need to replace `“` and `”` .

computermacgyver: I also think we may have the wrong character in the first entry? We're missing the single closing smart quote: `’`

skyemeedan (commented):
I think this would be easier to implement if there are tests added that cover the substitutions.  And this will save lots of time later as we add more cases to substitute

At a higher level,  do we know if the "sub keyword" examples we are seeing are coming from the same text (extracting both "Biden" and "Joe Biden") which hopefully will be well handled by this approach  vs multiple texts (one article referring to "Biden" and another to "President Biden") which might need some other post-processing

lib/model/yake_keywords.py
ahmednasserswe: I am not sure if the suggested implementation will cover all the cases.

I've added unit tests for this 
```
        keywords_test = [('Alegre',0), ('AlegreAlegre', 0), ('Alegre Alegre', 0), ("Presto", 0)]
        expected = [('AlegreAlegre', 0), ('Alegre Alegre', 0), ('Presto', 0)]
```

ahmednasserswe: Added more test cases 
        keywords_test = [('Alegre', 0),('Alegre', 0),('Timpani', 0), ('Presto Timpani', 0), ('AlegreAlegre', 0), ('Alegre Alegre', 0), ("Presto", 0)]
        expected = [('Presto Timpani', 0), ('AlegreAlegre', 0), ('Alegre Alegre', 0)]

lib/model/yake_keywords.py
```diff
@@ -26,15 +43,31 @@ def run_yake(self, text: str,
         :param num_of_keywords: int
         :returns: str
         """
+        ### if language is set to "auto", auto-detect it.
```
skyemeedan, line 46: I still think we need unit tests for the replacement/substitution part (because this list is likely to grow a lot).  so maybe worth a tiny refactor to put this part in its own function so easy to test

skyemeedan (approved):

lib/model/yake_keywords.py
```diff
@@ -7,8 +7,36 @@
...
+            if keep_keyword:
+                cleaned_keywords.append(keywords[i])
+        return cleaned_keywords
+    def normalize_special_characters(self, text):
+        replacement = {"`": "'",
+                       "‘": "'",
+                       "’": "'",
+                       "“": "\"",
+                       "”": "\""}
+
```
computermacgyver, line 34: Looks good now. Thanks.

lib/model/yake_keywords.py
```diff
@@ -7,8 +7,36 @@
...
+                current_keyword = keywords[i][0]
+                other_keyword = keywords[j][0]
+                if len(other_keyword) > len(current_keyword):
+                    if other_keyword.find(current_keyword + " ") >= 0 or other_keyword.find(" " + current_keyword) >= 0:
+                        keep_keyword = False
+                        break
+            if keep_keyword:
+                cleaned_keywords.append(keywords[i])
+        return cleaned_keywords
+    def normalize_special_characters(self, text):
```
computermacgyver, line 28: minor: Make sure there is a blank line between functions per PEP / Python style guide.

lib/model/yake_keywords.py
```diff
@@ -26,15 +54,25 @@ def run_yake(self, text: str,
         :param num_of_keywords: int
         :returns: str
         """
+        ### if language is set to "auto", auto-detect it.
+        if language == 'auto':
+            language = cld3.get_language(text).language
```
computermacgyver, line 59: Do we need a `try...except` around this? Have you tried with unsupported languages / nonsense and ensured this doesn't blow up?

ahmednasserswe: I don't think we need try....except. I've tried with unsupported language Ewe and nonsense text. They both got a wrong language detected by cld3 but it didn't break. 
For Yake, if you give it a wrong language or nonsupported one or you write something nonsense like "sdadsd" as the language code, the performance decreases but it doesn't break.

computermacgyver (approved):

DGaffney (approved):

PR #108, opened 2024-08-19
CV2-5101 i think you can just.... delete the subclass function and it'll just work?

DGaffney:
## Description
Brings text errors back in line with other fingerprinting errors

Reference: CV2-5101

## How has this been tested?
Not yet tested locally

No dependency change

No security change.

### DIFF
--- a/lib/model/generic_transformer.py
+++ b/lib/model/generic_transformer.py
@@ -75,10 +75,3 @@ class GenericTransformerModel(Model):
         Vectorize the text! Run as batch.
         """
         return self.model.encode(texts).tolist()
-
-    def handle_fingerprinting_error(self, error: Exception):
-        """
-        Handle any error that occurs during vectorization.
-        """
-        logger.error(f"Error during vectorization: {error}")
-        raise error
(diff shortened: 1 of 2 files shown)


lib/model/generic_transformer.py
```diff
@@ -75,10 +75,3 @@ def vectorize(self, texts: List[str]) -> List[List[float]]:
         Vectorize the text! Run as batch.
         """
         return self.model.encode(texts).tolist()
-
```
ashkankzme, line 78: perfect. i just took a look at this file. and maybe im just not fully understanding what this module does, so please let me know if I'm missing something.

it seems that you override `respond()` , which is the function that calls `get_response()` in `model.py`
`get_response()` handles caching and errors. 

i see that the overrided `respond()` takes care of caching, but not error handling. what would happen if an exception is thrown here? I guess it would still be logged but will be counted as a presto level error?

what if we try/catch exceptions in `respond()` and invoke `handle_fingerprinting_error()` in case of exceptions? there won't be a need to override `handle_fingerprinting_error()` either.

however if you think it will be fine as is, we can do this later too. I already see that we may have to refactor this file anyway with the new validation functionality coming in.

ashkankzme: nvm, Devin just walked me through the code and this comment is addressed

DGaffney: RESOLVED IN DMS

ashkankzme (approved):

PR #111, opened 2024-09-24
[CV2-5117] batch formats

skyemeedan:
## Description
Proposal for changes and batchifycation of presto request and response structures, standardization of argument handling across classes, and some suggestions on edits to get us there

Reference: CV2-5117

## How has this been tested?

### DIFF
--- a/docs/request.response.structures.md
+++ b/docs/request.response.structures.md
@@ -1,0 +1,185 @@
+# Expected incoming request and outgoing response datastructures for Presto models
+
+Presto models should use these structures to ensure that tools that consume the data can access in a consistant way across models
+
+Details of specific input formats are usually in the schema.py fiels
+
+## Incoming requests to presto
+
+General design
+
+* Models that don't handle requests in parallel need to implement batch format (but can give errors if they recieve more than 1 item)
+* "top level" elements are processed and controlled by the Presto system
+* elements inside parameters, and inside individual input items are passed through to the model
+* All of the content in a batch (a single html request) must go to the same model
+* Any parameter settings to the model must apply to all items in batch
+* Request itself needs a unique id, (and this is different than the id of the content)
+* There is single callback URL for entire batch
+* Some items need additional input parameters per item (i.e. “language” for YAKE), Presto just passes thes through
+
+Example PROPOSED input object structure (classycat)
+```
+{
+    request_id:122,              
+    model_name:"classycat__Model",       # model to be used 
+    model_parameters: {
+        "event_type": "classify",
+        "schema_id": "4a026b82-4a16-440d-aed7-bec07af12205",
+    },
+    input_items:[
+        {
+            id:"11",
+            text:"this is some text to classify", 
+            workspace_id:"timpani_meedan", 
+            target_state:"classified", 
+        },
+        {
+            id:"12",
+            text:"Este é um texto para classificar", 
+            workspace_id:"timpani_meedan", 
+            target_state:"classified", 
+        },
+    ],
+    callback_url: "http://conductor/batch_reply/",
+}
+
+```
+Example PROPOSED input object structure (YAKE)
+{
+    request_id:422,              
+    model_name:"yake_keywords__Model",       # model to be used 
+    model_parameters: {
+        "max_ngram_size":3,
+        "deduplication_threshold":0.25
+    },
+    input_items:[
+        {
+            id:"111",
+            text:"this is some text to classify", 
+            workspace_id:"timpani_meedan", 
+            target_state:"keyworded", 
+            language:"en",
+
+        },
+        {
+            id:"112",
+            text:"Este é um texto para classificar", 
+            workspace_id:"timpani_meedan", 
+            target_state:"keyworded", 
+            language:"pt"
+        },
+    ],
+    callback_url: "http://conductor/batch_reply/",
+}
+
+
+Example PROPOSED input object structure (paraphrase-multilingual)
+
+
+
+Example 'curl' command
+
+
+## Outgoing requests from presto
+
+* The outgoing reponse request includes the payloads and parameters from the incoming request
+* items in the response must to include their id, so that the caller can match individual-level properties from incoming request. 
+* status code is present at top level so it can be checked to know how to parse the rest of the 
+
+Example PROPOSED output structure
+```
+{
+    request_id:122,              
+    model_name:"classycat__Model",       # model to be used 
+    model_parameters: {
+        "event_type": "classify",
+        "schema_id": "4a026b82-4a16-440d-aed7-bec07af12205",
+    },
+    input_items:[
+        {
+            id:"11",
+            text:"this is some text to classify", 
+            workspace_id:"timpani_meedan", 
+            target_state:"classified", 
+            language:"en"
+
+        },
+        {
+            id:"12",
+            text:"Este é um texto para classificar", 
+            workspace_id:"timpani_meedan", 
+            target_state:"classified", 
+            language:"pt"
+        },
+    ],
+    callback_url: "http://conductor/batch_reply/",
+
+    # elements below here are added in response
+
+    "retry_count": 0
+    "status_code": 200
+    "status_message":"success"
+    "results": [
+            {
+                "id": "11",     #  this id can be used to look up params from input
+                "text": "this is some text to classify",   # text is redundant with input, but helpful for model-level debugging
+                "labels": [
+                    "Testing",
+                    "Classification"
+                ]
+            },
+            {
+                "id": "12",
+                "text": "Este é um texto para classificar",
+                "labels": [
+                    "Testing",
+                    "Classification"
+                ]
+            }
+        ]
+    },
+
+    ```
+
+Example PROPOSED output object structure (YAKE)
+{
+    request_id:422,              
+    model_name:"yake_keywords__Model",       # model to be used 
+    model_parameters: {
+        "max_ngram_size":3,
+        "deduplication_threshold":0.25
+    },
+    input_items:[
+        {
+            id:"11",
+            text:"this is some text to classify", 
+            workspace_id:"timpani_meedan", 
+            target_state:"keyworded", 
+            language:"en",
+
+        },
+    ],
+    callback_url: "http://conductor/batch_reply/",
+    "retry_count": 0
+    "status_code": 200
+    "status_message":"success"
+    "results": [
+            {
+                "id": "11",     #  this id can be used to look up params from input
+                "text": "this is some text to classify",  
+                "keywords": [
+                    ["lookup",0.0020211625251083634],
+                    ["params",0.0020211625251083634]
+                ]
+            },
+           
+        ]
+}
+
+# Error response structures
+
+"status_code": 403
+"status_message":"unable to access model"
+
+## Not covered
+per item errors?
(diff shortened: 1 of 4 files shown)


lib/model/classycat_classify.py
```diff
@@ -23,50 +23,55 @@ def get_client(self):
     def classify(self, task_prompt, items_count, max_tokens_per_item=200):
         pass
 
+
```
skyemeedan, line 26: sorry, I think all changes in this file are just updates from formatter

lib/model/yake_keywords.py
```diff
@@ -53,15 +56,28 @@ def run_yake(self, text: str,
         :param num_of_keywords: int
         :returns: str
         """
+        # TODO: loop over passed in items, or actually trigger processing in batch?
```
DGaffney, line 59: This looks fine as a first pass on the idea imo

lib/schemas.py
```diff
@@ -23,59 +35,77 @@ class VideoResponse(MediaResponse):
...
+    raw: Optional[Dict] = {}  # TODO: is this needed?
+    parameters: Optional[Dict] = (
+        {}
+    )  # optional parameters specific to processing this item (i.e. language)
 
 
 class Message(BaseModel):
-    body: GenericItem
+    request_id: Union[str, int, float]
+    items: List[GenericItem]  # to support batch, this is a list (can be only 1 item)
```
DGaffney, line 55: Looks fine to me, not terribly complicated!

lib/schemas.py
```diff
@@ -23,59 +35,77 @@ class VideoResponse(MediaResponse):
...
-class YakeKeywordsResponse(BaseModel):
+class YakeKeywordsResponse(ResponseItem):
     keywords: Optional[List[List[Union[str, float]]]] = None
 
 
 class GenericItem(BaseModel):
-    id: Union[str, int, float]
+    id: Union[str, int, float]  # id (in calling system) for this content
     content_hash: Optional[str] = None
-    callback_url: Optional[str] = None
```
DGaffney, line 33: Taking this out of here feels bad - why couldn't each item have its own callback?

skyemeedan: Classycat is implemented so it is making a single callback with all of the data in it.   I suppose we could support both options?  Like if the callback is present at the top-level, make a single callback with all the results,  if at individual level, respond per item.?(but does the entire batch input get returned with each individual item? Or presto has to parse apart the input list)  I guess having neither or both types of callback would be an error.

skyemeedan: Thinking about this further, I really like having the callback at the upper level, because it is an instruction to Presto system, not to the model (model doesn't know what to do with it)

DGaffney: That's fine, but it does likely mean that we can't use batch processing for any event on Check API without some somewhat significant revisions - we'd likely be adding callback functionality specifically to deal with batching. Not a terrible outcome, but does make it more complicated elsehwere

skyemeedan: > batch processing for any event on Check API without some somewhat significant revisions

Can you explain how you see this working so we can ensure the format encompasses that as well?

DGaffney: Sure - right now, if we allowed for the callback urls at the individual item level, we'd basically not have to make any changes to Check-API (or Alegre for that matter). The downside of course is that you'd open yourself up to making a ton of HTTP calls, which, not great (IMO i think the answer here is to allow callbacks at *either* level), but also, no revisions and potentially singular paths for receiving responses for work regardless of if the provenance was from a batch job or an individual job. If we go the route of only a top-level callback, we'd probably have to do some temporary redis key storage to remember what was sent out, and what needs to be done upon receiving, the results of work. Also, if we're running large batches of vectored work.... are we signing up for potentially gigantic POST bodies? Will they work in SQS even, or do they need to even live in SQS? I think they would?

Anyways to answer your question I think the meat of the issue would be that we'd have to introduce a bunch more state management paths for remembering and dealing with the outcomes of callbacks on Check API and Alegre - right now it's basically implied because the structure of the request contains all we need to know, but potentially just introduces more overhead. Not a terrible thing! Just something to consider.

skyemeedan: OK, so it sounds like for the format, you'd be ok with "callback can be at individual item or batch level (but must be at least one and not both)"?

> remember what was sent out, and what needs to be done upon receiving, the results of work. 

Or, you have to carry enough state along with the object to know what to do when it comes back.  Which is what we are trying to support in making sure the ids match back to the original payload.  ie. for timpani, the items carry a `target_state`.

lib/model/classycat_classify.py
computermacgyver: I link line 292 is the only substantive change: `for item in data["input_items"]:`

lib/schemas.py
computermacgyver: I'm generally in the callback belongs at the model level and not at the item level camp. 

I think it's right that we would make changes to Alegre/Check API if they start sending batches of items. * If a caller wants individual callbacks, they should send items one at a time. 
* If the caller wants one callback for a batch, then they should send a batch of items.

Right now, everything on Alegre/Check is set for sending items one at a time. When we implement the bulk similarity endpoint on Alegre, I'd expect to make changes to the callback handling in Alegre so that it can receive back one batch of items. This makes sense because Alegre can then process that batch efficiently as a batch --- i.e., calling the bulk insertion point on Open/Elasticsearch. Having items comeback one at a time from a bulk submission prevents the caller from handling the response as a batch.

tl;dr fine if we want to support callbacks at an individual level, but I don't see it as a priority because we already have a workaround: If a caller wants individual callbacks from Presto they should send items to Presto one at a time.

DGaffney: Alright I'm bought in for the toplevel callback - thank you both for the discussion, I think it was useful to at least air it out as it'll have ripple effects. Ship it at the top level!

lib/model/classycat_response.py
```diff
@@ -3,11 +3,12 @@
 
 
 class ClassyCatResponse(BaseModel):
+    # TODO: replase with the presto repose class?
```
ashkankzme, line 6: is this to replace `ClassyCatResponse` with presto response class? can you elaborate a bit more on what this could be like in practice please?

lib/model/yake_keywords.py
ashkankzme: @ahmednasserswe adding you here for visibility, let us know if you have any thoughts about this file specifically or the refactor as a whole.

lib/schemas.py
```diff
@@ -23,59 +35,77 @@ class VideoResponse(MediaResponse):
...
+    )  # assumes input is valid
 
     # TODO: the following is a temporary solution to handle the case where the model does not have a
     # parse_input_message method implemented but we must ultimately implement parse_input_message and
     # validate_input in all models. ticket: https://meedan.atlassian.net/browse/CV2-5093
-    if result_instance is None:  # in case the model does not have a parse_input_message method implemented
-        if 'yake_keywords' in model_name:
+    if (
+        result_instance is None
+    ):  # in case the model does not have a parse_input_message method implemented
```
ashkankzme, line 99: imo this is not a more readable formatting than what was before lol, any chance to manually make this pretty or revert this line back? 

@skyemeedan is there a way to opt out of formatting everything in the future? aside from cases like these, it makes it difficult to find out what the actual diffs are.

ashkankzme (changes requested):
sorry that it took me a while to get to it. I tried my best to be thorough and read everything in depth and left some comments. looking forward to the discussions!

docs/request.response.structures.md
```diff
@@ -0,0 +1,178 @@
...
+Details of specific input formats are usually in the schema.py fiels
+
+## Incoming requests to presto
+
+General design
+
+* Models that don't handle requests in parallel need to implement batch format (but can give errors if they recieve more than 1 item)
+* "top level" elements are processed and controlled by the Presto system
+* Elements inside parameters, and inside individual input items are passed through to the model
+* All of the content in a batch (a single html request) must go to the same model
```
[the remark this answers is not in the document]
skyemeedan: I guess I mean more precisely, "from the perspective of the calling service, all the of the items in a batch must be dispatched to the same Presto model service".   Classycat can of course internally decide to process via different LLMs "under the hood"  but don't mix payloads intended for classycat with payloads intended for paraphrase multilingual in the same call.

docs/request.response.structures.md
```diff
@@ -0,0 +1,178 @@
...
+
+Details of specific input formats are usually in the schema.py fiels
+
+## Incoming requests to presto
+
+General design
+
+* Models that don't handle requests in parallel need to implement batch format (but can give errors if they recieve more than 1 item)
+* "top level" elements are processed and controlled by the Presto system
+* Elements inside parameters, and inside individual input items are passed through to the model
```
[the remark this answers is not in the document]
skyemeedan: yeah, this is confusing.  By "top level" I mean the outer-most set of dictionary keys in the payload.  I think everything is passed through to the model?  But Presto is responsible for enforcing and checking keys at the outermost level, but shouldn't really enforce or check inside parameters or items lists (other than that ids exist).   Anything that isn't in the presto schema isn't promised to be passed through (we could always add to the schema if it becomes too constraining).  

I think greater flexibility is often faster in the short term but then ends up with everything having to do lots of conditional checking about elements that may or may not exist vs being able to check and enforce schemas and fail back to caller before enqueuing malformed messages to models

docs/request.response.structures.md
```diff
@@ -0,0 +1,178 @@
...
+Presto models should use these structures to ensure that tools that consume the data can access in a consistant way across models
+
+Details of specific input formats are usually in the schema.py fiels
+
+## Incoming requests to presto
+
+General design
+
+* Models that don't handle requests in parallel need to implement batch format (but can give errors if they recieve more than 1 item)
+* "top level" elements are processed and controlled by the Presto system
```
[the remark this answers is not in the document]
skyemeedan: yes, it is specific fields (which we should name) and the general concept that the outermost set of keys talk to Presto, and the next level down talk to the specific presto  model service.

lib/model/classycat_response.py
skyemeedan: lol this comment may beat my personal best for number of spelling errors per line!

I think I was meaning that the response message and list of items that ClassyCat was defining are now part of the Presto`Response class in schemas.py and enforced for all models.  So ClassyCat can import and extend that class and the ResponseItem class (into the classy schemas files) to add the ClassyCat specific fields

lib/schemas.py
skyemeedan: I agree it is hard to read when substantive and formatting changes are mixed together.  Which is why I'm a huge fan of configuring my editor to use a strict PEP-8 syntax formatting hook (especially if we all do it!)   I find it saves a huge amount of time  not having to manually mess with formatting or linting fixes as much.  

I could make a separate PR to first apply the formatting to the whole repo? ;-)  
I can also just move the comment up a line and it should fit within the line length

lib/model/classycat_classify.py
```diff
@@ -233,19 +273,23 @@ def process(self, message: Message) -> ClassyCatBatchClassificationResponse:
...
 
-        if "items" not in data["parameters"] or len(data["parameters"]["items"]) == 0:
-            raise PrestoBaseException("items are required as input to classify", 422)
+        if "input_items" not in data or len(data["input_items"]) == 0:
+            raise PrestoBaseException(
+                "input items are required as input to classify", 422
+            )
 
-        for item in data["parameters"]["items"]:
+        for item in data["input_items"]:
```
skyemeedan, line 292: As scott notes, this is the substantive change in this file,  now looping over the set of input items, unnested from `parameters`

PR #112, opened 2024-09-24
CV2-5247-support-yake-keyword-extraction-for-chinese

ahmednasserswe:
## Description
using jieba to segment mandarin text

Reference: CV2-5247 and CV2-4909
## How has this been tested?
locally. will add automated tests as well. 

### DIFF
--- a/lib/model/yake_keywords.py
+++ b/lib/model/yake_keywords.py
@@ -8,6 +8,7 @@ from lib import schemas
 
 import yake
 import cld3
+import jieba
 
 class Model(Model):
 
@@ -58,6 +59,9 @@ class Model(Model):
             language = cld3.get_language(text).language
         ### normalize special characters
         text = self.normalize_special_characters(text)
+        # Segmentation for mandarin
+        if language == 'zh-CN' or language == 'zh' or language == 'zh-TW':
+            text = " ".join(list(jieba.cut_for_search(text)))
         ### extract keywords
         custom_kw_extractor = yake.KeywordExtractor(lan=language, n=max_ngram_size, dedupLim=deduplication_threshold,
                                                     dedupFunc=deduplication_algo, windowsSize=window_size,
--- a/lib/model/yake_keywords.py
+++ b/lib/model/yake_keywords.py
@@ -39,6 +39,7 @@ class Model(Model):
 
     def run_chinese_segmentation_with_jieba(self, text):
         return " ".join(list(jieba.cut_for_search(text)))
+    
     def run_yake(self, text: str,
                  language: str,
                  max_ngram_size: int,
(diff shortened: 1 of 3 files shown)


skyemeedan (changes requested):
Can you add a test with a text example that shows this working?  Ideally would be a chunk of non space delimited Chinese text that is badly parsed without jieba and then returns appropriate keywords when processed with jieba.

skyemeedan (approved):
seems like a really cool library!

lib/model/yake_keywords.py
```diff
@@ -36,6 +37,8 @@ def normalize_special_characters(self, text):
             text = text.replace(k, v)
         return text
 
+    def run_chinese_segmentation_with_jieba(self, text):
+        return " ".join(list(jieba.cut_for_search(text)))
```
DGaffney, line 41: add newline after function please

DGaffney:
One extremely minor note otherwise good to go!

DGaffney (approved):

lib/model/yake_keywords.py
```diff
@@ -58,6 +61,10 @@ def run_yake(self, text: str,
             language = cld3.get_language(text).language
         ### normalize special characters
         text = self.normalize_special_characters(text)
+        # Segmentation for mandarin
+        if language == 'zh-CN' or language == 'zh' or language == 'zh-TW':
```
computermacgyver, line 65: Let's change to `if language[:2]=="zh"`

lib/model/yake_keywords.py
computermacgyver: I don't see a need for this to be a separate function given it's one line and only invoked in one place. I'd just include `text = " ".join(list(jieba.cut_for_search(text)))` right in the `run_yake` function

lib/model/yake_keywords.py
```diff
@@ -58,6 +61,10 @@ def run_yake(self, text: str,
...
+            text = self.run_chinese_segmentation_with_jieba(text)
+            # text = " ".join(list(jieba.cut_for_search(text)))
```
computermacgyver, line 67: Per above, I would drop the function call and just use `text = " ".join(list(jieba.cut_for_search(text)))` here. If we do stick with the function call, then please remove the commented code.

lib/model/yake_keywords.py
ahmednasserswe: I wanted to put it in a separate function so I can write tests separately for it. namely `test_run_chinese_segmentation_with_jieba` As this step includes downloading and loading a model. Is that the clean way to do so?

lib/model/yake_keywords.py
ahmednasserswe: added my comment about this above. please, Let me know what do you think about it.

computermacgyver: Sounds good. Let's leave as is.

PR #114, opened 2024-10-10
Using Spacy instead of Jieba for chinese text segmentation

ahmednasserswe:
## Description
Using Spacy instead of Jieba for Chinese text segmentation

Reference: CV2-5247

## How has this been tested?
Locally and added unittests

Spacy

### DIFF
--- a/test/lib/model/test_yake_keywords.py
+++ b/test/lib/model/test_yake_keywords.py
@@ -53,10 +53,10 @@ class TestYakeKeywordsModel(unittest.TestCase):
         expected = "'''\"\""
         self.assertEqual(self.yake_model.normalize_special_characters(text), expected)
 
-    def test_run_chinese_segmentation_with_jieba(self):
+    def test_run_chinese_segmentation(self):
         test_text = '''哈里斯同意与特朗普再进行一次美大选辩论'''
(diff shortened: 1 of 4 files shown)


test/lib/model/test_yake_keywords.py
```diff
@@ -53,10 +53,10 @@ def test_normalize_special_characters(self):
         expected = "'''\"\""
         self.assertEqual(self.yake_model.normalize_special_characters(text), expected)
 
-    def test_run_chinese_segmentation_with_jieba(self):
+    def test_run_chinese_segmentation(self):
         test_text = '''哈里斯同意与特朗普再进行一次美大选辩论'''
-        expected = "哈里 里斯 哈里斯 同意 与 特朗普 再 进行 一次 美 大选 辩论"
-        self.assertEqual(self.yake_model.run_chinese_segmentation_with_jieba(test_text), expected)
+        expected = "哈里斯 同意 与 特朗 普再 进行 一 次 美 大选 辩论"
```
skyemeedan, line 58: can we add tests with some of the examples I provided where jieba was giving multiple terms?

skyemeedan:
The code for switching over looks fine, but I wouldn't want to merge it in yet unless we've done some sort of comparison (perhaps you already did?).   We could dump some example text from piyaoba workspace and compare results side by side?

PR #116, opened 2024-10-21
[CV2-5117] spec doc describing batch format only

skyemeedan:
## Description

This splits off only the documentation of the presto batch refactor from https://github.com/meedan/presto/pull/111. so that we can go ahead and merge it separately to close https://meedan.atlassian.net/browse/CV2-5117

Actually Implementing the refactor for yake and classycat will be another ticket

### DIFF
--- a/docs/request.response.structures.md
+++ b/docs/request.response.structures.md
@@ -1,0 +1,179 @@
+# Expected incoming request and outgoing response datastructures for Presto models
+
+Presto models should use these structures to ensure that tools that consume the data can access in a consistant way across models
+
+Details of specific model's input and response formats are usually in the `<model_name>_response.py` files.  i.e. `classycat_response.py`
+
+## Incoming requests to presto
+
+General design
+
+* Models that don't handle requests in parallel need to implement batch format (but can give errors if they recieve more than 1 item)
+* "top level" elements in the payload (outermost dictionary keys) are processed and controlled by the Presto system
+* Elements inside parameters, and inside individual input items are passed through to the model
+* From the perspective of the calling service, all the of the items in a single batch call must be dispatched to the same Presto model service.  (don't mix items for paraphrase and classycat it the same call)
+* Any parameter settings to the model must apply to all items in batch
+* Request itself needs a unique id, (and this is different than the id of the content)
+* There is single callback URL for entire batch
+* Some items need additional input parameters per item (i.e. “language” for YAKE), Presto just passes thes through
+* The input structure will be returned in the response by the presto system, so individual models don't have to copy inputs to outputs. As long as ..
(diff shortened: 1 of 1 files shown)


skyemeedan:
This is the same document reviewed in previous PR, but without the refactor

docs/request.response.structures.md
```diff
@@ -0,0 +1,179 @@
...
+                "keywords": [
+                    ["lookup",0.0020211625251083634],
+                    ["params",0.0020211625251083634]
+                ]
+            },
+           
+        ]
+}
+
+# Error response structures
```
ashkankzme, line 173: it would be helpful if you could elaborate more on how we want to use the http error codes in responses. if an example helps, classycat returns a whole range of different error codes that would be good for explaining/reviewing a variety of use cases.

docs/request.response.structures.md
```diff
@@ -0,0 +1,179 @@
...
+
+Presto models should use these structures to ensure that tools that consume the data can access in a consistant way across models
+
+Details of specific model's input and response formats are usually in the `<model_name>_response.py` files.  i.e. `classycat_response.py`
+
+## Incoming requests to presto
+
+General design
+
+* Models that don't handle requests in parallel need to implement batch format (but can give errors if they recieve more than 1 item)
```
DGaffney, line 11: I think all models should support batch format, and those that can't do parallel ops just do them serially then return, IMO...

docs/request.response.structures.md
```diff
@@ -0,0 +1,179 @@
...
+* "top level" elements in the payload (outermost dictionary keys) are processed and controlled by the Presto system
+* Elements inside parameters, and inside individual input items are passed through to the model
+* From the perspective of the calling service, all the of the items in a single batch call must be dispatched to the same Presto model service.  (don't mix items for paraphrase and classycat it the same call)
```
DGaffney, line 14: Hard agree on this assumption

docs/request.response.structures.md
```diff
@@ -0,0 +1,179 @@
...
+* Any parameter settings to the model must apply to all items in batch
```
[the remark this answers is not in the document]
DGaffney: My question here would be how would global parameters even be honored / used? Should all that info just get delegated down to the per-item level of detail? Could get messy to have that info at the task level

docs/request.response.structures.md
```diff
@@ -0,0 +1,179 @@
...
+* Request itself needs a unique id, (and this is different than the id of the content)
```
DGaffney, line 16: Does it though? I could see not providing it and just getting a job id back that's a randomly assigned UUID on the response from presto?

PR #120, opened 2024-11-13
refactor to get rid of persistent queues and change to singular lookup

DGaffney:
## Description
Queues could theoretically get wonky with long-held state, and also we could be more specific when selecting queues by using singular access instead of filters.

Reference: CV2-5589

## How has this been tested?
Not tested locally yet

Nothing changes in this PR

Nothing changes in this PR


sonoransun (approved):
Looks good!

skyemeedan:
does it not break when run locally with elasticmq?

PR #121, opened 2024-11-14
CV2-5589 More support for queuing under threaded environments

DGaffney:
## Description
Responding to more issues where we fail to pickup queues under threaded environments. More work to make it threadsafe

Reference: CV2-5589

## How has this been tested?
Tested locally and continues to work!

No change

No Change

### DIFF
--- a/lib/queue/queue.py
+++ b/lib/queue/queue.py
@@ -111,10 +111,10 @@ class Queue:
         Retrieve or create a queue with the specified name.
         """
         try:
-            return self.get_sqs().get_queue_by_name(QueueName=queue_name)
+            return [self.get_sqs().get_queue_by_name(QueueName=queue_name)]
         except botocore.exceptions.ClientError as e:
             if e.response['Error']['Code'] == "AWS.SimpleQueueService.NonExistentQueue":
-                return self.create_queue(queue_name)
+                return [self.create_queue(queue_name)]
             else:
                 raise
 
--- a/lib/queue/queue.py
+++ b/lib/queue/queue.py
@@ -118,16 +118,6 @@ class Queue:
             else:
                 raise
 
-    def create_queue(self, queue_name: str):
-        """
-        Create a queue by name.
-        """
-        attributes = {}
-        if queue_name.endswith('.fifo'):
-            attributes['FifoQueue'] = 'true'
-            attributes['ContentBasedDeduplication'] = 'true'
-        return self.get_sqs().create_queue(QueueName=queue_name, Attributes=attributes)
-
     def send_message(self, queue_name: str, message: schemas.Message):
         """
         Send a message to a specific queue.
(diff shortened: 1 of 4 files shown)


sonoransun (approved):
Thread local sqs objects should hopefully do it! 🙏

lib/queue/queue.py
```diff
@@ -98,27 +118,29 @@ def get_or_create_queues(self, queue_name: str) -> List[boto3.resources.base.Ser
...
-                                      aws_access_key_id=(get_environment_setting("AWS_ACCESS_KEY_ID") or 'x'),
-                                      aws_secret_access_key=(get_environment_setting("AWS_SECRET_ACCESS_KEY") or 'x'))
-        else:
-            logger.info(f"Using SQS Interface")
-            with self.sqs_lock:
-                return boto3.resource('sqs', region_name=get_environment_setting("AWS_DEFAULT_REGION"))
+        attributes = {}
+        if queue_name.endswith('.fifo'):
+            attributes['FifoQueue'] = 'true'
+            attributes['ContentBasedDeduplication'] = 'true'
```
computermacgyver, line 128: Is attributes['ContentBasedDeduplication'] related to the queue ending in .fifo or is it something we want all the time?

lib/queue/queue.py
```diff
@@ -98,27 +118,29 @@ def get_or_create_queues(self, queue_name: str) -> List[boto3.resources.base.Ser
...
+        return self.get_sqs().create_queue(QueueName=queue_name, Attributes=attributes)
+
+    def send_message(self, queue_name: str, message: schemas.Message):
+        """
+        Send a message to a specific queue.
+        """
+        queue = self.get_or_create_queue(queue_name)
```
computermacgyver, line 135: Do we really need to look up the queue each time we send a message? I would have hoped we could store a reference after looking it up once

computermacgyver (approved):

lib/queue/queue.py
DGaffney: My understanding is that if we _do_ store it we introduce a bunch of state which causes our problem in the first place

sonoransun: The first call it will create the resource object, but then it will return a thread local reference on every subsequent request (within that thread). I think this is necessary due to the context switching that could happen otherwise...

lib/queue/queue.py
DGaffney: its a fifo thing - and I realized I copypasta'd here twice so removing one of these!

sonoransun: Right; the deduplication is an aspect of how FIFO queues work (where you avoid processing the same message more than once). I don't think we'll ever have a scenario with FIFO, but without deduplication, but it doesn't seem unreasonable to separate these aspects.

PR #122, opened 2024-11-14
CV2-5589 fix bad variable name

DGaffney:
## Description
Stray variable, bad name - easy fix!

Reference: CV2-5589

## How has this been tested?
Not tested locally, confirmed error

No changes

No changes

### DIFF
--- a/lib/queue/queue.py
+++ b/lib/queue/queue.py
@@ -169,7 +169,7 @@ class Queue:
         """
         Receive messages from a queue.
         """
-        queue = self.get_or_create_queue(self.input_queue)
+        queue = self.get_or_create_queue(self.input_queue_name)
         return queue.receive_messages(MaxNumberOfMessages=min(batch_size, SQS_MAX_BATCH_SIZE))
 
     def find_queue_by_name(self, queue_name: str) -> boto3.resources.base.ServiceResource:


sonoransun (approved):
✔️

computermacgyver (approved):

skyemeedan:
If neither our local testing or CI test on https://github.com/meedan/presto/commit/a91954a39bdf644948d448e1deb7e4502f367585  was able to catch this error that the code did not run at all, then I think it needs to be a priority to make sure we have testing that can cover this kind of thing (smoke test? integration test?).  It may take a day or so to write but will probably greatly speed up development process!

DGaffney:
> If neither our local testing or CI test on [a91954a](https://github.com/meedan/presto/commit/a91954a39bdf644948d448e1deb7e4502f367585) was able to catch this error that the code did not run at all, then I think it needs to be a priority to make sure we have testing that can cover this kind of thing (smoke test? integration test?). It may take a day or so to write but will probably greatly speed up development process!

Agreed - I do not have the bandwidth to get it in immediately, do you want to put another set of eyes on this code and see if you can smoke it out? I would love to get some knowledge overlap generally since I have a high bus factor on this code as it stands

PR #124, opened 2024-11-15
CV2-5589 reverts accidental regression

DGaffney:
## Description
Reverts regression on structure of receive_messages which returns tuples, and changed as a result of moving to singular queue responses

Reference: CV2-5589

## How has this been tested?
Tests confirm this was a regression - other tests really should have caught this outside just the singular case

None

None


sonoransun (approved):
Looks good 🙏

computermacgyver (approved):

PR #125, opened 2024-11-15
CV2-5589 resolve broken requests to delete messages

DGaffney:
## Description
More knock-ons of bad test mocks and switchover from multi stateful to single stateless

Reference: CV2-5589

## How has this been tested?
Worked through the current issue on sentry locally and confirmed this will return a queue properly

None

None


sonoransun (approved):
✔️

PR #128, opened 2024-11-19
CV2-5709 readd http condition only for non-workers

DGaffney:
## Description
Re-adding http after tests confirming http seems to thrash the box

Reference: CV2-5709

## How has this been tested?
None

None

None


sonoransun (approved):
Danke sir!

PR #133, opened 2024-11-26
CV2-5726 don't use reload flag when running the server in the cloud

DGaffney:
## Description
We use `--reload` when we really shouldn't in production on Presto - it slows requests down and the server will, by definition, never reload.

Reference: CV2-5726

## How has this been tested?
Not tested locally, by definition, since this applies to prod

None

None


skyemeedan (approved):
This seems like a great change!  I'm assuming we will know right away in QA if the performance changed oddly

PR #134, opened 2024-12-02
CV2-5639 add sleep during message receives

DGaffney:
## Description
We hemorrhage a ton of `EmptyReceives` when requesting messages via SQS - this PR slows down the requestor to an amount that is negligible for us, but will greatly reduce ridiculous levels of needless spend.

Reference: CV2-5639

## How has this been tested?
By its nature this is a remote/cloud issue

none

none


sonoransun (approved):
✔️

PR #138, opened 2025-01-03
[CV2-5925] standardize messages for logs

DGaffney:
## Description
When you run `logger.error` in the context of an exception, that percolates to sentry as an error with whatever the `logger.error` text is. This converts those cases to use our capture message format instead to sync them all up.

Reference: CV2-5925

## How has this been tested?
Not yet tested locally, CI should discover any issues here.

None

Not relevant

### DIFF
--- a/lib/queue/processor.py
+++ b/lib/queue/processor.py
@@ -6,7 +6,7 @@ import requests
 from lib import schemas
 from lib.logger import logger
 from lib.queue.queue import Queue
-
+from lib.sentry import capture_custom_message
 
 class QueueProcessor(Queue):
     @classmethod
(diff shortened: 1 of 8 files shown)


lib/queue/processor.py
```diff
@@ -68,9 +68,7 @@ def send_callback(self, message):
                 # headers={"Content-Type": "application/json"},
             )
             if response.ok != True:
-                logger.error(f"Callback error responding to {callback_url} :{response}")
+                capture_custom_message("Callback response not ok", 'error', {"response": response, "callback_url": callback_url})
```
skyemeedan, line 71: can we include the response error code as well?  this would probably help a lot in debuging and appropriately grouping errors

skyemeedan (approved):

